PR #1, opened 2016-12-15
Abstract Node over some monad

avieth:
The initial sketch was all `IO`. Now, it's any `m`, subject to typeclass constraints depending upon which non-pure functionality the program uses.

`Network/Transport/Abstract.hs` was copied, trimmed, and abstracted from the real `network-transport` package. Multicast stuff was removed. The types which didn't have to change have been recycled.

I've also given a way to use a concrete network transport like tcp in the abstracted one. See `Network/Transport/Concrete.hs`.

Listener dispatching is partially implemented, and I've included a working ping/pong example.


avieth:
@ChShersh @dcoutts

src/Mockable/SharedAtomic.hs
```diff
@@ -0,0 +1,35 @@
...
+
+data SharedAtomic (m :: * -> *) (t :: *) where
+    NewSharedAtomic :: t -> SharedAtomic m (SharedAtomicT m t)
+    ModifySharedAtomic :: SharedAtomicT m s -> (s -> m (s, t)) -> SharedAtomic m t
+
+newSharedAtomic :: ( Mockable SharedAtomic m ) => t -> m (SharedAtomicT m t)
+newSharedAtomic t = liftMockable $ NewSharedAtomic t
+
+readSharedAtomic :: ( Mockable SharedAtomic m ) => SharedAtomicT m t -> m t
+readSharedAtomic sat = modifySharedAtomic sat (\x -> pure (x, x))
```
ChShersh: Just wonder: why implement `read` in terms of `modify`? For me it seems more reasonable to implement `modify` using `read` and `write`. Implementing `read` using `modify` may introduce some unnecessary overhead.

src/Network/Transport/Concrete.hs
```diff
@@ -0,0 +1,56 @@
...
+concrete :: MonadIO m => NT.Transport -> Transport m
+concrete transport = Transport {
+      newEndPoint = concreteNewEndPoint (NT.newEndPoint transport)
+    , closeTransport = liftIO $ NT.closeTransport transport
+    }
+
+concreteNewEndPoint
+    :: ( MonadIO m )
+    => IO (Either (TransportError NewEndPointErrorCode) NT.EndPoint)
+    -> m (Either (TransportError NewEndPointErrorCode) (EndPoint m))
```
ChShersh: Maybe it's better to use `ExceptT` here in future?

src/Node.hs
```diff
@@ -8,33 +8,53 @@
...
+
+nodeId :: Node m -> NodeId
+nodeId = LL.NodeId . NT.address . LL.nodeEndPoint . nodeLL
+
 type Worker m = SendActions m -> m ()
 
-type MessageName = BS.ByteString
+newtype MessageName = MessageName BS.ByteString
+deriving instance Eq MessageName
+deriving instance Ord MessageName
```
ChShersh: Just curious: why `StandaloneDeriving` instead of `GeneralizedNewtypeDeriving` here?

src/Node.hs
```diff
@@ -69,17 +89,28 @@ data ConversationActions m = ConversationActions {
...
+       , MonadFix m )
     => NT.Transport m
     -> g
     -> [Worker m]
     -> [Listener m]
     -> m (Node m)
-startNode transport prng workers listeners = do
-    node <- LL.startNode transport prng handlerIn handlerInOut
+startNode transport prng workers listeners = mdo
+    node <- LL.startNode transport prng (handlerIn node sendAction) (handlerInOut node)
```
ChShersh: Why do we need to use `RecursiveDo` here and pass `node` in `hadlerInOut` if this `node` isn't used in that handler now? Or we just probably may want to use it there?

src/Node.hs
```diff
@@ -89,27 +120,62 @@ startNode transport prng workers listeners = do
...
+        let listener = M.lookup msgName listenerIndex
+        case listener of
+            Just (ListenerActionOneMsg action) -> do
+                (msgBody, rest') <- recvPart chan rest
+                tid <- fork (action peerId sendActions msgBody)
+                -- TODO remember the thread id? Inform dispatcher when it's
+                -- finished?
+                pure ()
+            -- If it's a conversation listener, then that's an error, no?
+            Just (ListenerActionConversation _) -> error ("Wrong listener type! " ++ show msgName)
```
ChShersh: Seems to me like it's better to use `throw` here since we have `Mockable Throw m` constraint.
But yes, I agree, that here should be an error if we encounter conversation listener according to my understanding of model.

src/Node.hs
```diff
@@ -69,17 +89,28 @@ data ConversationActions m = ConversationActions {
...
     => NT.Transport m
     -> g
     -> [Worker m]
     -> [Listener m]
     -> m (Node m)
-startNode transport prng workers listeners = do
-    node <- LL.startNode transport prng handlerIn handlerInOut
+startNode transport prng workers listeners = mdo
+    node <- LL.startNode transport prng (handlerIn node sendAction) (handlerInOut node)
     let sendAction = SendActions { sendTo = sendMsg node, connect = undefined }
```
ChShersh: Don't really get what should be here later instead of `undefined`... :confused:

src/Node.hs
avieth: I just prefer standalone deriving syntax. I could do it in the typical way instead if you like.

src/Node.hs
avieth: I'm working on that now, but will merge it in another pull request.

src/Node.hs
ChShersh: This is absolutely minor and not requires rewritings) We usually prefer `GND` because it's shorter to write.

PR #4, opened 2016-12-20
Avieth/discovery

georgeee:
This includes a 'discovery' abstraction and an implementation of it with Kademlia (demonstration in `examples/Discovery.hs`, as well as support for bidirectional connections (demonstration in `examples/PingPong.hs`).

### DIFF
--- a/examples/PingPong.hs
+++ b/examples/PingPong.hs
@@ -3,17 +3,25 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE RecursiveDo #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE StandaloneDeriving #-}
 
+import GHC.Generics (Generic)
 import Control.Monad (forM_)
 import Data.String (fromString)
+import Data.Proxy (Proxy(..))
+import Data.Void (Void)
+import Data.Binary
 import Node
 import qualified Network.Transport.TCP as TCP
 import qualified Network.Transport.InMemory as InMemory
 import Network.Transport.Concrete (concrete)
 import System.Random
 import qualified Control.Concurrent as Conc
-import qualified Control.Concurrent.Chan as Conc
 import qualified Control.Concurrent.MVar as Conc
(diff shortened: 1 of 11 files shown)


examples/PingPong.hs
```diff
@@ -37,38 +45,79 @@ instance Mockable SharedAtomic IO where
...
+instance Mockable Catch IO where
+    liftMockable (Catch action handler) = action `Exception.catch` handler
+
+-- Sending a message which encodes to "" is problematic!
+-- The receiver can't distinuish this from the case in which the sender sent
+-- nothing at all.
+-- So we give custom Ping and Pong types with non-generic Binary instances.
+--
+-- TBD should we fix this in network-transport? Maybe every chunk is prefixed
+-- by a byte giving its length? Wasteful I guess but maybe not a problem.
```
avieth: @dcoutts This annoys me but it's probably not a big deal. I wanted to use `()` as the pong message type, but it serializes to `""` and so the programs in `Node` and `Node.Internal` can't distinguish a sent `()` from an end-of-input (or from two or more sent `()`s). If network-transport preserved the sender's chunking then it would be OK; we'd get a [""] and push that onto the in-channel. But maybe this comes with too high a price (extra data on the wire).

examples/PingPong.hs
```diff
@@ -37,38 +45,79 @@ instance Mockable SharedAtomic IO where
...
-            putStrLn (show id ++ " is waiting for " ++ show i ++ "us before sending pings")
+            --putStrLn (show id ++ " is waiting for " ++ show i ++ "us before sending pings")
             Conc.threadDelay i
-            forM_ peerIds (\peerId -> sendTo sendActions peerId (fromString "ping") ())
+            let pong :: NodeId -> ConversationActions Void Pong IO -> IO ()
+                pong peerId cactions = do
+                    putStrLn (show id ++ " sent PING to " ++ show peerId)
+                    Pong <- recv cactions
+                    putStrLn (show id ++ " heard PONG from " ++ show peerId)
+            forM_ peerIds $ \peerId -> withConnectionTo sendActions peerId (fromString "ping") (pong peerId)
```
avieth: Notice that the pinger never `send`s anything. The `MessageName` sent to the peer (`"ping"`) essentially fills the role of a send.

PR #6, opened 2016-12-27
[] Little additions in README and PingPong example.

denisshevchenko:
It's not about some task, it's just about little improvements.


denisshevchenko:
I've added `Ping` type (because I see notice about Ping in the comments, but only Pong type was defined). And small overview in the `README.md`.

PR #7, opened 2017-01-03
Concatenate chunks before attempting to parse

avieth:
This is the one true way. See the discussion here:
https://github.com/haskell-distributed/network-transport-tcp/pull/44

network-transport only guarantees that
`concat sentChunks == concat receivedChunks`
which neatly handles the case of `[]` as `concat [] = ""`.

Types which serialize to `""` are now acceptable.


avieth:
Why is master merged into this request before the request is merged back?

ChShersh:
@avieth When I tried to accept PR some conflicts appeared. So when I resolved them via Github webpage `master` was merged into this PR. Don't know if it can be avoided using only this frontend (not terminal one).

PR #8, opened 2017-01-05
Feature/csl447 new tw timed

denisshevchenko:
Fix warnings and corrections in examples (no more `delay`, it's `wait`).


examples/Discovery.hs
```diff
@@ -30,55 +31,56 @@ import           System.Random
 data Pong = Pong
 deriving instance Show Pong
 instance Binary Pong where
-    put _ = putWord8 (fromIntegral 1)
+    put _ = putWord8 $ fromIntegral (1 :: Int)
```
georgeee: `$ fromIntegral (1 :: Int)` could be replaced by `1`, not?

PR #9, opened 2017-01-05
Add Async mocking; change node start/stop; some questions about time/delay mocking

avieth:
- `Async` mocking has been added. @georgeee has requested adding `mapConcurrently`. That can certainly be done and will be useful.
- `startNode` has been removed in favour of a function called `node` which doesn't spawn workers for you, but gives you a `Node m` and `SendActions m` to work with, shutting down the `Node m` automatically when your work is done.
- The dependency on `time-warp` has been removed. Surely we don't want this to depend on `time-warp`, since it's a rework of that package.
- I have some concerns about `repeatForever` and `sleepForever`. They can both be implemented in a simpler way in terms of other primitives, and I don't think `sleepForever` is a good idea at all. Let's discuss.

### DIFF
--- a/src/Mockable/Concurrent.hs
+++ b/src/Mockable/Concurrent.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
 
 module Mockable.Concurrent (
 
@@ -24,6 +25,11 @@ module Mockable.Concurrent (
   , async
   , wait
   , cancel
+
+  , Concurrently(..)
+  , concurrently
+  , mapConcurrently
+  , forConcurrently
 
   ) where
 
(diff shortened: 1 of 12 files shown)


georgeee:
Can you add `mapConcurrently` via your async?

avieth:
`mapConcurrently` and `forConcurrently` are included.

Test suite now compiles and goes. There was a bug in the node dispatcher, with regard to conversation handshaking. If node `A` starts a conversation with node `B` but node `A` never `recv`s any data, and its handler finishes, before node `B`'s ACK has been received, the dispatcher would throw a protocol error. But in the case just described, it's not an error.

Martoon-00:
Great, thanks!

src/Mockable/Concurrent.hs
```diff
@@ -50,29 +47,16 @@ myThreadId = liftMockable MyThreadId
 killThread :: ( Mockable Fork m ) => ThreadId m -> m ()
 killThread tid = liftMockable $ KillThread tid
 
+type RelativeToNow = Microsecond -> Microsecond
```
Martoon-00: I guess we don't need this and `for` anymore, can just make `delay :: TimeUnit t => t -> m ()`

georgeee: Nice idea

PR #10, opened 2017-01-05
Sketch of Kademlia-based broadcasting

avieth:
By using Kademlia's network overlay, we can broadcast to the network without flooding (explanation below).

`examples/Broadcast.hs` demonstrates that it works. I have run it for 80 nodes locally, 20 broadcasts, and it's consistent (every node receives every message).

Please give comments on the interface for broadcasting. I'm not so happy with it but it's a good start I think.

## Broadcasting using Kademlia's overlay

Every node in the network is a Kademlia peer, and so has an `n`-bit identifier (typical case is `n = 160`). These identifiers give the network a tree structure with every node as a leaf. Find its position by interpreting a `1` bit as "go left" and a `0` bit as "go right" so that each `n`-bit identifier gives a path of length `n` from root to leaf.

When a node wishes to *initiate* a broadcast, it sends the message to one peer from each *bucket* (definition later) that it's aware of, and it adds to the data payload its own identifier. When a node wishes to *repeat* a broadcast that it heard from some other peer, it uses that peer identifier which it received with the data, along with its own identifier, to determine which nodes to repeat the message to, such that they certainly have not already been contacted. It sends the same payload to them, but with its own identifier instead of the sender's.

What are these buckets? Every node is a leaf in Kademlia's routing tree, so imagine walking back up that path from node to root. At each level, there's a subtree which contains the node, and another subtree which does not. The latter subtree is what I call a bucket. The broadcast initiator sends the message to one node from each of these buckets. Those nodes, if they wish to repeat the broadcast, will send the message to one node from each of *their* buckets, restricted to those nodes which are closer to them than to the sender, i.e. they are, from the sender's point of view, within the same bucket as the receiver.

See section 4.3.2 of [this paper](http://www.hiradastechnika.hu/data/upload/file/2009/2009%20I/Pages_from_HT0901a_2.pdf).

### DIFF
--- a/examples/Broadcast.hs
+++ b/examples/Broadcast.hs
@@ -111,10 +111,12 @@ makeBroadcaster transport n = do
         pure ()
     worker discovery bcast@(initiator, _) prng n sactions = do
         let (m :: Int, prng') = random prng
+        liftIO . putStrLn $ "Discovering"
         _ <- discoverPeers (K.kdDiscovery discovery)
         peers <- knownPeers (K.kdDiscovery discovery)
+        liftIO . putStrLn $ "Broadcasting " ++ show m
         initiator m sactions
-        delay (for 500000)
+        delay (for 5000000)
         worker discovery bcast prng' (n - 1) sactions
 
 forK :: [t] -> (t -> m () -> m ()) -> m () -> m ()
--- a/examples/Broadcast.hs
+++ b/examples/Broadcast.hs
@@ -1,0 +1,154 @@
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE GADTs                 #-}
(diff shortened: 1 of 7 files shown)


jagajaga:
ping @georgeee @ChShersh @Martoon-00 @dcoutts

georgeee:
This feature was implemented by Alex Vieth on his initiative and actually we don't need this feature now. SO it's left as PR for inclusion in future if we would need it

avieth:
> This feature was implemented by Alex Vieth on his initiative and actually we don't need this feature now. SO it's left as PR for inclusion in future if we would need it

I believe it's already been merged... this stuff is available in `master`. Also note that the merge target is yet another branch that has already been merged, `avieth/async`.

avieth:
Broken branch. Closing but not abandoning. We may want a broadcasting abstraction some day, but not now.

PR #13, opened 2017-01-11
Preliminary example of delaying network input

avieth:
This is just a demo of my progress towards being able to selectively delay input from a given peer.

It requires a patch to network-transport-tcp, as shown in the diff of `stack.yaml`. That has complicated the use of a concrete network-transport, because the patch only applies to network-transport-tcp, not to every network-transport. How or if this feature would be included in network-transport is up for debate.

Hard-coded into `Node.Internal` is a policy which delays every socket read from every peer by 5 milliseconds. To make a useful backpressure feature we'll allow the programmer to specify how to compute such a delay for each peer from a set of statistics.

### DIFF
--- a/examples/Discovery.hs
+++ b/examples/Discovery.hs
@@ -24,7 +24,7 @@ import           Mockable.Production
 import           Network.Discovery.Abstract
 import qualified Network.Discovery.Transport.Kademlia as K
 import           Network.Transport.Abstract           (Transport (..))
-import           Network.Transport.Concrete           (concrete)
+import           Network.Transport.Concrete.TCP       (concreteTCP)
 import qualified Network.Transport.TCP                as TCP
 import           Node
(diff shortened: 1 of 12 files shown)


georgeee:
I reviewed changes. I don't really get
1. What this delay policy is actually used for?
2. And when, on which condition? Is it some mocking-related behavior? Seems like not

avieth:
> What this delay policy is actually used for?

It's how we specify if and for how long we delay taking input from a given peer.

> And when, on which condition? Is it some mocking-related behavior? Seems like not

Of course, this is not a complete feature, just a preliminary demo of the infrastructure on which the feature could be built.

> To make a useful backpressure feature we'll allow the programmer to specify how to compute such a delay for each peer from a set of statistics.

avieth:
Closing in favour of #16

PR #14, opened 2017-01-14
Add logging

Martoon-00:


src/Node/Internal.hs
```diff
@@ -354,24 +463,26 @@ nodeDispatcher endpoint nodeState handlerIn handlerInOut =
...
                   case outcome of
                       Nothing -> pure Nothing
-                      Just (tid, ChannelIn chan) -> do
+                      Just (promise, ChannelIn chan) -> do
                           Channel.writeChannel chan (Just (BS.concat (LBS.toChunks ws')))
-                          pure . Just $ ConnectionReceiving tid chan
+                          pure . Just $ ConnectionReceiving promise (ChannelIn chan)
 
                 | otherwise ->
                     throw (ProtocolError $ "unexpected control header " ++ show w)
```
avieth: Log don't throw.

src/Node/Internal.hs
```diff
@@ -383,31 +494,48 @@ nodeDispatcher endpoint nodeState handlerIn handlerInOut =
...
-                  Just (ConnectionNewChunks peer@(NodeId _) chunks0) -> do
-                      mConnState <- handleFirstChunks connid peer (chunks0 ++ chunks) state
-                      loop (maybe state (\cs -> Map.insert connid cs state) mConnState)
+                  Just (peer, ConnectionNew) ->
+                      loop $ state {
+                            dispatcherConnections = Map.insert connid (peer, ConnectionNewChunks chunks) (dispatcherConnections state)
+                          }
+
+                  Just (peer, ConnectionNewChunks chunks0) -> do
+                      emConnState <- try $ handleFirstChunks connid peer (chunks0 ++ chunks) state
```
avieth: Skeptical of using `try` here. Exceptions in `handleFirstChunks` should be re-thrown I think.

src/Node/Internal.hs
```diff
@@ -417,71 +545,147 @@ nodeDispatcher endpoint nodeState handlerIn handlerInOut =
...
+                            dispatcherConnections = Map.insert connid (peer, ConnectionClosed promise) (dispatcherConnections state)
+                          }
 
                   -- Small message
-                  Just (ConnectionNewChunks peer chunks0) -> do
-                      mConnState <- handleFirstChunks connid peer chunks0 state
-                      case mConnState of
-                          Just (ConnectionReceiving tid chan) -> do
+                  Just (peer, ConnectionNewChunks chunks0) -> do
+                      emConnState <- try $ handleFirstChunks connid peer chunks0 state
```
avieth: See previous comment about use of `try` on `handleFirstChunks`.

src/Node/Internal.hs
Martoon-00: AFAIU if I rethrow, then dispatcher loop would stop. Is it okay?

avieth: Yes I think this is what we want. If there's an exception then we really do want to know about it.

Martoon-00:
Okay, I'm going to apply your notes, thanks for review!

PR #15, opened 2017-01-16
Cl integration

avieth:


src/Node.hs
```diff
@@ -98,7 +100,7 @@ listenerMessageName (ListenerActionConversation f) =
                 -> Proxy rcv
                 -> MessageName
         msgName _ = messageName
-    in  msgName (f undefined) Proxy
+    in  msgName (f undefined undefined) Proxy
```
avieth: What's with the `undefined`s? Is this safe?

src/Node.hs
```diff
@@ -77,14 +77,16 @@ data ListenerAction packing m where
   -- | A listener that handles an incoming bi-directional conversation.
   ListenerActionConversation
     :: ( Packable packing snd, Unpackable packing rcv, Message rcv )
-    => (LL.NodeId -> ConversationActions snd rcv m -> m ())
```
avieth: I'm skeptical of this because it gives great power to conversation listeners. Should they be able to send outside of the conversation itself? Start new conversations? Is there a particular use case that motivated this change?

bench/Sender/Main.hs
```diff
@@ -56,7 +57,7 @@ main = do
     loadLogConfig logsPrefix logConfig
     setLocaleEncoding utf8
 
-    Right transport_ <- TCP.createTransport "0.0.0.0" "3432" TCP.defaultTCPParameters
+    Right transport_ <- TCP.createTransport "0.0.0.0" "127.0.0.1" "3432" TCP.defaultTCPParameters
```
avieth: I'm confused about this. Why is it not OK to bind on `127.0.0.1`?

georgeee: Probably it's ok for tests and benchs.

georgeee: Or better said certainly it's ok. I just didn't think much

src/Node.hs
georgeee: Probably better replaced with Proxy constructors. But given we never actually pattern-match on proxy, these are kind of equal solutions

src/Node.hs
georgeee: Suppose we exchanged some data and then want to distribute it. Though probably this should be done with workers... We had one case actually, but then it was reworked to worker :)

src/Node.hs
Martoon-00: Should be safe since `msgName`'s arguments is not used. @avieth if you consider it bad style, I'll remove these `undefined`s

src/Node.hs
Martoon-00: Yeah, and then we had no time to rollback till now. @georgeee do you think some use case may appear again, or we'd better rollback?

bench/Sender/Main.hs
avieth: I think we ought to give only one address: the bind address. We can give `0.0.0.0`, and then patch nt-tcp so that, when a connection comes it, it uses `getSocketName` to determine the address at which the peer reached us (it won't be `0.0.0.0`).

src/Node.hs
georgeee: Well. Maybe usecase will arise. But as I understand now, it really brings lot more of responsibility providing interface to use `sendActions`. So we better use solution with workers and rollback code. Also I would prefer to strengthen this (not using sendActions from conversation) by logging error if use being detected (that is not hard to do with thread-local variables)

src/Node.hs
georgeee: @Martoon-00 Better use `Proxy` constructor, vote for this as well

avieth: I think we can avoid the scary `undefined`s and do this using scoped type variables. Give a type signature for `f`, i.e. `LL.NodeId -> SendActions packing m -> ConversationActions snd rcv m -> m ()`, so that `rcv` is in scope. Then use `messageName (Proxy :: Proxy rcv)`.

bench/Sender/Main.hs
georgeee: Sounds good. I thought about this aready, but not familiar enough with nt-tcp internal code to do this fix

avieth: Oh, I see. Took me a while to realize what the issue is. `address :: EndPoint -> EndPointAddress` will be `0.0.0.0:<port>:<id>` if we bind to `0.0.0.0`, but of course we don't want to tell our peers to connect to this address.

avieth: Perhaps it would be best to leave nt-tcp unpatched and instead take the bind address of the node as a config parameter. Binding on `0.0.0.0` will not be allowed, and you'll have to know your public address.

georgeee: @avieth binding on oublic ip night not work on some systems. E.g. if you don;t have interface with this ip on your system, like we encountered on AWS

PR #16, opened 2017-01-19
Backpressure

avieth:
The current implementation of our node has a single dispatcher thread, pulling network-transport `Event`s from an unbounded `Chan` and then determining what to do with them, typically putting received data into some other unbounded `Chan`. That's no good. network-transport-tcp will pull data from sockets as fast as it can, no matter what.

Even with queueing and buffering for network-transport-tcp as in https://github.com/haskell-distributed/network-transport-tcp/pull/46 we're still not quite there. We could define some `QDisc` which bounds the input buffers for each socket, but we're still left with only one means of output: `receive :: EndPoint -> Event`. It aggregates events from *every* socket, and then our dispatcher still hands out the data to the unbounded `Chan`s of the handlers.

So I've made a somewhat radical change to network-transport-tcp which you can find here: https://github.com/avieth/network-transport-tcp/tree/avieth/backpressure . It allows for the `QDisc` to give out a custom type, not necessarily `Event`. This makes the `ConnectionBuffers :: QDisc EndPointEvent` possible (see `src/Network/Transport/ConnectionBuffers.hs`). This puts out an event for every peer connection which includes a bounded input buffer for that connection alone. The main dispatcher thread of our node can spin as fast as it wants, picking up new connections and handing off their input buffers to handlers. Those handlers will be responsible for reading from the buffers. A full buffer leads to a slowdown on the peer's side, hence we have very natural, somewhat automatic backpressure: if your handler is slow, it'll slow the input from the peer.

Also included in this patch is some statistics-gathering in `Node.Internal`. Although they're currently not used to inform a backpressure policy, they *are* used in the abuse example to feed an EKG server. You can run that example and then view live statistics by opening `localhost:8000`:

```
./examples/abuse/Main server 7777 +RTS -N1 -T -RTS
# Now open the EKG console at localhost:8000 and view the flat graphs.
./examples/abuse/Main client 7777 7778 +RTS -N -RTS
# Watch the graphs in the EKG console as the server is spammed with data
```

### DIFF
--- a/bench/Receiver/Main.hs
+++ b/bench/Receiver/Main.hs
@@ -10,7 +10,6 @@ import           Control.Monad              (unless)
 
 import           Data.Time.Units            (Second)
 import           GHC.IO.Encoding            (setLocaleEncoding, utf8)
-import qualified Network.Transport.TCP      as TCP
 import           Options.Applicative.Simple (simpleOptions)
 import           Serokell.Util.Concurrent   (threadDelay)
 import           System.Random              (mkStdGen)
@@ -20,7 +19,8 @@ import           Mockable                   (Production (runProduction))
 
 import           Bench.Network.Commons      (MeasureEvent (..), Ping (..), Pong (..),
                                              loadLogConfig, logMeasure)
-import           Network.Transport.Concrete (concrete)
+import qualified Network.Transport.TCP      as TCP
+import qualified Network.Transport.Concrete.TCP as TCP
 import           Node                       (ListenerAction (..), NodeAction (..), node,
                                              sendTo)
 import           Node.Message               (BinaryP (..))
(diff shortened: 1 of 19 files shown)


avieth:
Dropping lightweight connections isn't possible in nt-tcp at present. If there's already a heavyweight connection, then no handshake takes place for new connections; a control header is sent and then the lightweight connection is assumed ready, the peer has no chance to reject it (the lightweight connection only, leaving the socket and its other lightweight connections in tact).

Suppose some peer has induced an inordinate amount of handlers in our node. Say there are `n` handlers currently running in response to that peer, and `n` is something like 4 standard deviations from the mean number of handlers per peer. What can we do in response to this? We want to keep taking in data from the peer, for the handlers may still need some and we must not starve them. But we don't want to take this new connection, because we want to maintain high quality of service for other peers. The only reasonable option I see is to drop it.

We could either:

1. Impose a handshake on new lightweight connections, so that the abusive peer will see an error when trying to connect to the overloaded peer who dropped his connection.
2. Silently drop new data from that peer.

Number 1 isn't so bad. The cost of the handshake won't be high, the socket is already up. Number 2 could be bad: if the peer tries to get a bidirectional connection going, it will time out waiting for the ACK; if the peer tries to start a unidirectional connection, he'll never know that the peer ignored him. Is that ok? Currently when we send a unidirectional message we have no guarantee that the peer did anything with it anyway.

PR #17, opened 2017-01-26
Node carries static peer data and transmits on new connections

avieth:
This is a requirement that was discussed with @georgeee 

We need to have application-specific, network-agnostic data about a peer, but it was determined that sending this information for every message would be too heavy. Another option was to have each node carry an application-specific identifier, and other nodes keep state associating those identifiers with this peer data, and I suppose having a way to demand peer data in case it's not known for that identifier.

The solution implemented here complicates the node quite a bit but it should be good. The first lightweight connection to a peer shall always be responsible for sending the peer data. Once the data is sent (`send` returns) other connections to that peer can begin sending data. The dispatcher therefore shall always know the peer data about every peer which has at least one connection to it before it performs a handshake or runs any application-specific handlers.

Also included in this changes is some metrics/statistics, and an abuse test case, from previous work.

### DIFF
--- a/bench/Receiver/Main.hs
+++ b/bench/Receiver/Main.hs
@@ -10,7 +10,6 @@ import           Control.Monad              (unless)
 
 import           Data.Time.Units            (Second)
 import           GHC.IO.Encoding            (setLocaleEncoding, utf8)
-import qualified Network.Transport.TCP      as TCP
 import           Options.Applicative.Simple (simpleOptions)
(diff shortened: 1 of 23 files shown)


avieth:
CI passes, that's great :) I've also observed a flat heap in the ping-pong and discovery examples. This one needs very careful review, though. Lots of potential for concurrency-related bugs.

PR #18, opened 2017-01-31
Improved abuse and backpressure demonstration

avieth:
See the [README](https://github.com/serokell/time-warp-nt/blob/ac0cf3c3c64b56b92ac3609fb0fa6d9215d3ee01/examples/abuse/README.md)

Also included in this PR are fixes for some mistakes which were uncovered while writing up documentation.

- The type of `Node.node` changes: the `NodeActions` (which includes the listeners) must be computed purely (rather than in `m`) and can use the components of the `Node` itself only lazily. That's because the listeners must be defined *prior* to the end point being brought up. Prior to this change there was a gap in time between bringing the end point up, and defining the listeners, so it was possible (though highly unlikely) that the system would try to run a listener which is not yet defined.
- The definition of `Node.ConversationActions` changes so that it no longer includes `peerData :: m peerData`. This was originally included so that the `ConversationActions` given to `withConnectionTo` would provide the initiating peer with the receiving peer's data once it is available (hence the `m peerData` rather than just `peerData`). However, `ConversationActions` is also given to a conversation listener, which already has the peer data available, so there was a confusing and unnecessary duplication there. Now, the peer data is given as another argument to the continuation of `withConnectionTo`.

### DIFF
--- a/bench/Receiver/Main.hs
+++ b/bench/Receiver/Main.hs
@@ -47,7 +47,7 @@ main = do
 
     runProduction $ usingLoggerName "receiver" $ do
         node transport prng BinaryP () $ \_ ->
-            pure $ NodeAction [pingListener noPong] $ \_ -> do
+            NodeAction [pingListener noPong] $ \_ -> do
                 threadDelay (fromIntegral duration :: Second)
   where
     pingListener noPong =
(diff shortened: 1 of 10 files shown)


georgeee:
Reviewed. To be merged to master with appropriate changes in CSL (later a bit)

avieth:
@georgeee rebased on master, tests/CI passes.

PR #20, opened 2017-02-06
[CSL-730] Catch&ignore exceptions from endPointClosed

volhovM:
https://issues.serokell.io/issue/CSL-730
cc @georgeee

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,10 @@
+with import <nixpkgs> { };
+
+haskell.lib.buildStackProject {
+   name = "node-sketch";
+   ghc = haskell.packages.ghc802.ghc;
+   buildInputs = [
+     openssh autoreconfHook git 
+   ] ++ (lib.optionals stdenv.isLinux [ cabal-install stack ]);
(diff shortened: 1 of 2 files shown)


src/Node/Internal.hs
```diff
@@ -623,7 +623,8 @@ nodeDispatcher node handlerIn handlerInOut =
           NT.ConnectionClosed connid -> connectionClosed state connid >>= loop
 
           -- When the end point closes, we're done.
-          NT.EndPointClosed -> endPointClosed state
+          NT.EndPointClosed -> 
+              endPointClosed state `catch` (\(e :: SomeException) -> pure ())
```
georgeee: Could you send errors to logDebug instead of just ignoring?

volhovM:
I've removed two commits because the same thing was implemented in the other branch. Though having `shell.nix` would be good.

avieth:
LGTM

avieth:
Looks like most of this was already merged, separately. Do we still need the remaining changes?

volhovM:
This was fixed already, PR is obsolete.

PR #21, opened 2017-02-07
Various bug fixes concerning outbound bidirectional connections

avieth:
- Use custom network-transport and network-transport-tcp branches which implement `ConnectionBundle` identifiers, and provide more regular `ConnectionClosed` events (for instance, if a connection is lost, all open lightweight connections are reported closed). Relevant nt issue https://github.com/haskell-distributed/network-transport/issues/33
- Sharpen exception reporting in node: exceptions in the action versus exceptions in trying to stop the node itself.
- Fix a bug in the dispatcher in which partial control headers would be incorrectly parsed.
- Eliminate races in connecting/disconnecting from a peer and determining which connection should send the peer data.
- Always plug the input channels on un-ACK'd locally-initiated conversation handlers.

### DIFF
--- a/node-sketch.cabal
+++ b/node-sketch.cabal
@@ -66,7 +66,6 @@ Library
                       , mtl
                       , network
                       , network-transport
-                      , network-transport-inmemory
                       , network-transport-tcp
                       , mtl >= 2.2.1
                       , random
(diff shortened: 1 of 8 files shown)


avieth:
@volhovM has requested that this be merged and I agree, it should be merged. Waiting for CI to pass and then I'll press the green button. CC @georgeee

PR #23, opened 2017-02-11
Improve testing, fix race on bidirectional connections

avieth:
The race was incredibly rare, probably never to be seen except on self-connections: it was possible to get the ACK before the handler/nonce was registered in the state.

Testing is improved: two sloppy new cases for self connections and peer data. Also updated the spec to use in-memory and TCP transports.

### DIFF
--- a/node-sketch.cabal
+++ b/node-sketch.cabal
@@ -227,7 +227,9 @@ test-suite node-sketch-test
                      , hspec >= 2.1.10
                      , lens >= 4.14
                      , mtl >= 2.2.1
+                     , network-transport
                      , network-transport-tcp
+                     , network-transport-inmemory
                      , node-sketch
                      , QuickCheck
                      , quickcheck-instances
--- a/src/Node/Internal.hs
+++ b/src/Node/Internal.hs
@@ -32,7 +32,7 @@ module Node.Internal (
   ) where
 
 import           Control.Exception             hiding (bracket, catch, finally, throw)
-import           Control.Monad                 (forM_, forM, when)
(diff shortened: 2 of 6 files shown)


georgeee:
Reviewed. Looks good

avieth:
One of the CI times out on the single-message test using TCP. Very strange. I'll look into this

avieth:
Hooray the CI passes. Today I found that the `push` version would always succeed, and the `pr` version would always fail. If I guess correctly, the former tests this branch itself, while the latter tests the merge of this branch into master. This puzzled me, because this branch has always been a fast-forward of master, i.e. the branch and its merge into master *are exactly the same* so why would one of the CIs consistently fail while the other consistently passes?

avieth:
> Hooray the CI passes. Today I found that the push version would always succeed, and the pr version would always fail. If I guess correctly, the former tests this branch itself, while the latter tests the merge of this branch into master. This puzzled me, because this branch has always been a fast-forward of master, i.e. the branch and its merge into master are exactly the same so why would one of the CIs consistently fail while the other consistently passes?

... and now it fails. But at least they both fail :)

Still frustrated by this. The logs of the failed tests strongly suggest a bug in nt-tcp, but I just can't reproduce it on my machine.

avieth:
Yes! Tracked down the reason for the test failures. This commit fixes them https://github.com/serokell/time-warp-nt/pull/23/commits/d2ee617eb8fb86df5decb94980454c3abc00d4af

It was a bug in network-transport-tcp. My patch is under review right now.

PR #24, opened 2017-02-14
[TW-132] Note on peer data and sessions

avieth:
View `PeerData.md` and add your comments/concerns here to the pull request.

### DIFF
--- a/notes/PeerData.md
+++ b/notes/PeerData.md
@@ -1,0 +1,192 @@
+# Static peer data
+
+## Inbound connections
+
+Currently we guarantee the peer data to be present for incoming connections:
+listeners and locally-initiated bidirectional connections. That's somewhat
+clear from the types:
+
+```Haskell
+peerData :: ConversationActions peerData body rcv m -> m peerData
+
+listenerActionOneMsg
+    :: peerData
+    -> NodeId
+    -> SendActions packing peerData m
+    -> msg
+    -> m ()
+
+-- Note the duplicate: peerData is here _and_ in the ConversationActions.
+-- This is fixed in one of my branches.
+listenerActionConversation
+    :: peerData
+    -> NodeId
+    -> ConversationActions peerData snd rcv m
+    -> m ()
+```
+
+The branch mentioned is [here](https://github.com/serokell/time-warp-nt/pull/18).
+
+time-warp-nt ensures that the peer data is not transmitted unnecessarily. *The
+peer data of a given peer is held for as long as there is at least one
+connection to or from it*. This makes sense, because the peer data is assumed
+to be constant for the lifetime of a node. If it is to change, the node must
+go down, and therefore all of its connections to any other node must be severed.
+
+## Outbound unidirectional connections
+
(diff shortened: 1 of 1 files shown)


georgeee:
Looks like having these sessions is pretty same in practice as having conversation with `Void` receive type. Hence I would prefer to avoid these changes and leave peer-data be available for inbound-only (+ out of conversation scope)

avieth:
> Looks like having these sessions is pretty same in practice as having conversation with Void receive type. Hence I would prefer to avoid these changes and leave peer-data be available for inbound-only (+ out of conversation scope)

I was under the impression that having peer data even for unidirectional outbound messages was a requirement. Is this not the case?

With sessions we can group outbound messages (whether unidirectional or bidirectional) to a certain peer so that they are all guaranteed to be sent to the same peer with the same peer data. Perhaps this won't be necessary? Can you imagine a case in which we need to group 2 or more unidirectional or bidirectional connections to a peer, with the guarantee that they reach the same peer running the same version? Currently we can get such a guarantee only for conversation-style messages.

This can also be somewhat of an optimization: by opening sessions with each of a node's neighbours, many messages can be sent and received with the guarantee that only one TCP connection will be established.

dcoutts:
> This can also be somewhat of an optimization: by opening sessions with each of a node's neighbours, many messages can be sent and received with the guarantee that only one TCP connection will be established.

I think strictly speaking this is an orthogonal feature, but a very useful feature in the same area of the code, and if we're doing both we could do both at the same time for efficiency.

I think the right design will be to use a dedicated keep-alive lightweight connection, and a one-off initial peer data connection, but that these things do not have to be directly related to a session that we expose in the time-warp API. A session at that level is just a guarantee that the peer doesn't change for the duration, but we don't have to pick a design where we set up and take down a connection with the peer for that session.

avieth:
@georgeee said today on Slack that this sounds reasonable. Perhaps we should schedule to do this work next week, as network-transport-tcp improvements are due by this Friday, March 3. In the mean-time we can of course keep discussing it. Do the proposed types look good?

avieth:
Holding open a lightweight connection, as far as I can tell, doesn't mean that the peer has not gone down and come up again. It does in the case of a TCP transport, but network-transport itself doesn't make this guarantee does it?

avieth:
This feature would help with support for unaddressable nodes (wallets behind NAT). Holding open a session with a peer makes a node addressable by that peer, even if the initiator's transport is not addressable. But as far as I know, a node contacts only peers that are judged to be its neighbours according to Kademlia. Unaddressable nodes can't be considered Kademlia peers. We would need a mechanism for responding to new sessions, so that we could record the `NodeId` of the connecting peer as a known peer (one which was not discovered by Kademlia). `sendToNeighbors`/`converseToNeighbors` could then include these peers as well.

avieth:
Any more questions/comments/concerns about this? I'm ready to go ahead with the interface given in the 'Sessions' section of the markdown file. @georgeee @dcoutts

PR #25, opened 2017-02-15
Merged master to cl-integration

Martoon-00:


avieth:
Is this something that we still need to do? Is cardano using this branch? I thought it was using master.

Martoon-00:
Yep, cardano uses `master` and we want to merge `cl-integration` -> `master`. But according to our workflow we should merge `master` -> `cl-integration` first  ¯\_(ツ)_/¯

avieth:
> Yep, cardano uses master and we want to merge cl-integration -> master. But according to our workflow we should merge master -> cl-integration first ¯_(ツ)_/¯

Seems they've diverged a lot. I tried a rebase of `cl-integration` onto `master` and it looks like many of the commits are either no longer necessary or must be redone, since `Node.Internal` has changed quite a bit. It may be best to simply go through the commits that are in `cl-integration` but not `master`, and cherry-pick the ones we want.

Martoon-00:
I did some `master` merges in `cl-integration`, they shouldn't actually differ much. I mean, if we merge, it should go smoothly

PR #26, opened 2017-02-16
Fair QDisc implementations and simulation

avieth:
Run the simulation like so:

  `stack exec qdisc-simulation -- <qdisc> <duration>`

where qdisc is one of
  "fair", "fair_stm", "one_place", "unbounded"
and duration parses to a whole number, determining for how many seconds
to run the simulation.
Yes there are two implementations of the fair QDisc. Seems they have
very close performance characteristics. Maybe we should favour the STM
variant because it's simpler.
You can change the distributions for reader and writers by modifying
Simulation.hs. What's in there now shows a case in which the fair QDisc
really is fairer than the one-place.

# Using a particular QDisc in the real-world

Currently, none of the time-warp examples nor the POS prototype actually use a non-default `QDisc`; they're all using the `simpleUnboundedQDisc` which never blocks on writes.

It may be a good idea now to swap in a different `QDisc`. Doing so is easy: when creating the TCP transport (`Network.Transport.TCP.createTransport`) supply your desired `QDisc`-maker as the `tcpNewQDisc` record field:

```Haskell
let tcpParameters = TCP.defaultTCPParameters { ... }
let myQDisc :: forall t . IO (QDisc t)
    myQDisc = simpleOnePlaceQDisc
createTransport bindTo address port (tcpParameteres { tcpNewQDisc = myQDisc })
```


avieth:
@dcoutts @kantp

avieth:
New commit removes the `MVar` variant in favour of `STM`. Also improves simulation output. Will merge once the CI goes green.

PR #27, opened 2017-02-28
Track total live bytes for each peer

avieth:
"live bytes" means those bytes which have been delivered to a handler
which is currently running. This supports a conservative assumption that
a handler uses memory proportional to the amount of bytes given to its
input channel.

The dispatcher thread is of course responsible for reporting the
increments, but `spawnHandler` does most of the work. Other programs can
inspect the node statistics, held in a `SharedAtomic`. The idea is that
a `QDisc` or perhaps the dispatcher thread itself will use this
information to induce an artificial delay on one or all peers in case of
heavy load (lots of live bytes).

### DIFF
--- a/src/Node/Internal.hs
+++ b/src/Node/Internal.hs
@@ -32,7 +32,7 @@ module Node.Internal (
   ) where
 
 import           Control.Exception             hiding (bracket, catch, finally, throw)
-import           Control.Monad                 (forM_, forM, when, unless)
+import           Control.Monad                 (forM_, forM, when)
 import           Control.Monad.Fix             (MonadFix)
 import           Data.Int                      (Int64)
 import           Data.Binary                   as Bin
@@ -217,7 +217,7 @@ data Statistics m = Statistics {
       --   locally, i.e. corresponding to bidirectional connections.
     , stRunningHandlersLocal :: !(Metrics.Gauge m)
       -- | Statistics for each peer.
-    , stPeerStatistics :: !(Map NT.EndPointAddress PeerStatistics)
+    , stPeerStatistics :: !(Map NT.EndPointAddress (SharedAtomicT m PeerStatistics))
       -- | How many peers are connected.
     , stPeers :: !(Metrics.Gauge m)
       -- | Average number of remotely-initiated handlers per peer.
@@ -236,11 +236,13 @@ data Statistics m = Statistics {
     , stHandlersFinishedExceptionally :: !(Metrics.Distribution m)
     }
 
-stTotalLiveBytes :: Statistics m -> Int
-stTotalLiveBytes stats = sum allBytes
-    where
-    allPeers = Map.toList (stPeerStatistics stats)
-    allBytes = fmap (pstLiveBytes . snd) allPeers
+stTotalLiveBytes
+    :: (Mockable SharedAtomic m)
+    => Statistics m -> m Int
+stTotalLiveBytes stats = do
+    allPeers <- mapM readSharedAtomic $ Map.elems (stPeerStatistics stats)
+    let allBytes = fmap (pstLiveBytes) allPeers
+    return $ sum allBytes
 
 stRunningHandlersRemoteVariance :: Statistics m -> Double
 stRunningHandlersRemoteVariance statistics = avg2 - (avg ^ 2)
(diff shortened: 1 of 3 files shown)


src/Node/Internal.hs
```diff
@@ -255,6 +261,9 @@ data PeerStatistics = PeerStatistics {
       -- | How many handlers are running right now for locally-iniaiated
       --   bidirectional connections to this peer.
     , pstRunningHandlersLocal :: !Int
+      -- | How many bytes have been received on open connections from this
+      --   peer.
```
avieth: Misleading comment. Will update it.

avieth:
@kantp

src/Node/Internal.hs
```diff
@@ -1141,32 +1176,38 @@ spawnHandler stateVar provenance action =
...
+                    alteration Nothing = Just $ Map.singleton nonce (someHandler, channelIn, incrBytes, peerDataVar, connBundle, False)
+                    alteration (Just map) = Just $ Map.insert nonce (someHandler, channelIn, incrBytes, peerDataVar, connBundle, False) map
                 Local peer Nothing -> nodeState {
                       _nodeStateOutboundUnidirectional = Map.insert peer someHandler (_nodeStateOutboundUnidirectional nodeState)
                     }
 
+            incrBytes !n = modifySharedAtomic stateVar $ \nodeState -> do
+                let statistics' = stIncrBytes (handlerProvenancePeer provenance) n (_nodeStateStatistics nodeState)
+                modifySharedAtomic totalBytes $ \(!m) -> return (m + n, ())
+                return (nodeState { _nodeStateStatistics = statistics' }, ())
```
avieth: `incrBytes` is what dispatcher will do when it receives data for a handler. It requires holding the node state open. I wonder if this will be an issue. There may be too much contention for this MVar.

kantp: Yes, that might become a point of contention.

As you suggested on slack, I'll change it so that the `Map` itself holds `MVar`s that can be modified each, without holding the `MVar` for the whole `Map`.

src/Node/Internal.hs
```diff
@@ -1141,32 +1192,38 @@ spawnHandler stateVar provenance action =
...
+                    alteration Nothing = Just $ Map.singleton nonce (someHandler, channelIn, incrBytes, peerDataVar, connBundle, False)
+                    alteration (Just map) = Just $ Map.insert nonce (someHandler, channelIn, incrBytes, peerDataVar, connBundle, False) map
                 Local peer Nothing -> nodeState {
                       _nodeStateOutboundUnidirectional = Map.insert peer someHandler (_nodeStateOutboundUnidirectional nodeState)
                     }
 
+            incrBytes !n = do
+                nodeState <- readSharedAtomic stateVar
+                stIncrBytes (handlerProvenancePeer provenance) n (_nodeStateStatistics nodeState)
+                modifySharedAtomic totalBytes $ \(!m) -> return (m + n, ())
```
avieth: `incrBytes` still contends for the `stateVar`. Perhaps `stAddHandler` should give back not just the new `Statistics` value, but also the `SharedAtomicT` of the live bytes for the peer.

kantp: But it's just reading the `stateVar` now, not taking it. That shouldn't be a problem for contention, should it?

avieth:
```Haskell
readMVar :: MVar a -> IO a
readMVar m =
  mask_ $ do
    a <- takeMVar m
    putMVar m a
    return a
```

kantp: Ah, not anymore :)

https://www.stackage.org/haddock/lts-8.2/base-4.9.1.0/src/GHC-MVar.html#readMVar

But it will be the same problem on older GHC versions. Can we live with that?

kantp: Oh, you're right!

We might want to change that, then.

avieth: > Ah, not anymore :)
>
> https://www.stackage.org/haddock/lts-8.2/base-4.9.1.0/src/GHC-MVar.html#readMVar
>
> But it will be the same problem on older GHC versions. Can we live with that?

Nice, wasn't aware of this. It's new in 4.9.1.0? I suppose it's not such a big deal if older versions have slightly worse performance, as long as they have the same semantics, which they do.

avieth: Ok I pushed a commit adding a `ReadSharedAtomic` constructor and using `readMVar` for `Production`.

PR #29, opened 2017-03-01
Update stack.yaml for nt-(tcp)(inmemory)

avieth:
A network-transport-tcp change for a separate bind address has
landed in master. The type of createTransport has changed: the third
argument takes the actual bound port and returns a host address and port
to use as the external address.

### DIFF
--- a/bench/Receiver/Main.hs
+++ b/bench/Receiver/Main.hs
@@ -39,7 +39,7 @@ main = do
     loadLogConfig logsPrefix logConfig
     setLocaleEncoding utf8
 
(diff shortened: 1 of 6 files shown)


avieth:
@kantp @georgeee 

Note the new type:

```Haskell
createTransport
  :: N.HostName    -- ^ Bind host name.
  -> N.ServiceName -- ^ Bind port.
  -> (N.ServiceName -> (N.HostName, N.ServiceName))
                   -- ^ External address host name and port, computed from the
                   --   actual bind port.
  -> TCPParameters
  -> IO (Either IOException Transport)
```

avieth:
I'm preparing a branch of `cardano-sl` which will work with these changes, then I'll merge.

PR #30, opened 2017-03-01
Fair QDisc can de-prioritize certain peers

avieth:
We may want to use this in conjunction with some mutable blacklist
determined by metrics gathered by a node, so as to induce artificial
delays on certain peers deemed to be abusive.

@kantp This might be useful in your abuse tests, if we can come
up with a suitable delay for a peer based upon the node's metrics
(for instance, using this feature https://github.com/serokell/time-warp-nt/pull/27).


kantp:
Thanks!

PR #31, opened 2017-03-02
SPECIALIZE INLINE Production mockables

avieth:
Also added -O2 GHC options to relevant modules.

@kantp


kantp:
Nice! Will we also want to add `INLINABLE` to functions like `modifySharedAtomic`?

avieth:
> Nice! Will we also want to add INLINABLE to functions like modifySharedAtomic?

Couldn't hurt. I'd hope GHC would figure this out given how small they are but I'll add them anyway.

kantp:
Great!

PR #32, opened 2017-03-03
[Tw 127] remove compiler warnings

kantp:
This reduces the noise at compile time, making it easier to spot warnings that might be important.

Also, replaced some incomplete pattern matches with Exceptions.

### DIFF
--- a/src/Node/Internal.hs
+++ b/src/Node/Internal.hs
@@ -397,7 +397,7 @@ stAddHandler !provenance !statistics = case provenance of
     -- TODO: generalize this computation so we can use the same thing for
     -- both local and remote. It's a copy/paste job right now swapping local
     -- for remote.
-    Local _ _ -> do
+    Local !_peer _ -> do
         (!peerStatistics, !isNewPeer) <- pstAddHandler provenance (stPeerStatistics statistics)
         when isNewPeer $ Metrics.incGauge (stPeers statistics)
         Metrics.incGauge (stRunningHandlersLocal statistics)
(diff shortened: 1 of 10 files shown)


src/Node/Internal.hs
```diff
@@ -412,7 +413,7 @@ stAddHandler !provenance !statistics = case provenance of
             , stRunningHandlersLocalAverage = runningHandlersLocalAverage
             }
 
-    Remote !peer _ _ -> do
+    Remote _ _ _ -> do
```
avieth: Should leave the bang `!_`

avieth: And on line 400 as well.

src/Node/Internal.hs
```diff
@@ -773,10 +774,10 @@ nodeDispatcher node handlerIn handlerInOut =
     connectionOpened
         :: DispatcherState peerData m
         -> NT.ConnectionId
-        -> NT.Reliability
+        -> NT.Reliability -- TODO: this is not actually used. If this is intentional, we should remove the parameter.
```
avieth: You can go ahead and remove it if you like. It's certainly not used now, we'll put it back if we ever need it.

avieth:
LGTM after those bang patterns are reinserted. Thanks @kantp

src/Node/Internal.hs
kantp: @avieth: also on lines 464 and 481?

avieth: Ah yes, good catch.

PR #33, opened 2017-03-07
[TW-114] timeout waiting for ACK

avieth:
A programmable timeout in microseconds is given when a node is created. If a bidirectional connection does not hear an ACK before this time then it's killed with a `Timeout` exception. A test case is included.

### DIFF
--- a/src/Node/Internal.hs
+++ b/src/Node/Internal.hs
@@ -18,6 +18,8 @@
 module Node.Internal (
     NodeId(..),
     Node(..),
+    NodeEnvironment(..),
+    defaultNodeEnvironment,
     nodeId,
     nodeEndPointAddress,
     Statistics(..),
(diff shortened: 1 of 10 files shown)


avieth:
I plan on doing some cleanup in here. The `_nodeStateOutboundBidirectional` type is unwieldly. I'll make it a record.

src/Node/Internal.hs
```diff
@@ -366,7 +386,7 @@ initialStatistics = do
 data HandlerProvenance peerData m t =
       -- | Initiated locally, _to_ this peer. The Nonce is present if and only
       --   if it's a bidirectional connection.
-      Local !NT.EndPointAddress (Maybe (Nonce, SharedExclusiveT m peerData, NT.ConnectionBundle, t))
+      Local !NT.EndPointAddress (Maybe (Nonce, SharedExclusiveT m peerData, NT.ConnectionBundle, Promise m (), t))
```
georgeee: I would really suggest using datatype here, guessing which meaning every element of tuple corresponds to is really cumbersome.

georgeee: Alright, created https://issues.serokell.io/issue/TW-133

PR #34, opened 2017-03-09
[TW-125] introduce rate limiting

kantp:
Adds rate-limiting, as a number of the in-flight bytes for a given peer.

### DIFF
--- a/src/Network/QDisc/Fair.hs
+++ b/src/Network/QDisc/Fair.hs
@@ -6,7 +6,6 @@
 module Network.QDisc.Fair (
 
       fairQDisc
-    , blockingQDisc
 
     ) where
 
@@ -15,16 +14,6 @@ import Control.Concurrent (threadDelay)
 import Network.Transport (EndPointAddress)
 import Network.Transport.TCP (QDisc(..))
 import Data.Time.Units (Microsecond)
-
-blockingQDisc :: (EndPointAddress -> IO ()) -> IO (QDisc t)
-blockingQDisc waitForUnblock = do
-    fqd <- newFairQDisc
-    return QDisc {
-          qdiscEnqueue = \addr _ t -> do
-                  waitForUnblock addr
-                  writeFairQDisc fqd t
-        , qdiscDequeue = readFairQDisc fqd
-        }
 
 -- | Make a fair 'QDisc'. It's called fair, but it can also be unfair if you
 --   want it to be, by having some 'EndPointAddress's delay before writing.
--- a/src/Network/QDisc/Fair.hs
+++ b/src/Network/QDisc/Fair.hs
@@ -6,6 +6,7 @@
 module Network.QDisc.Fair (
 
       fairQDisc
+    , blockingQDisc
 
     ) where
 
@@ -14,6 +15,16 @@ import Control.Concurrent (threadDelay)
 import Network.Transport (EndPointAddress)
 import Network.Transport.TCP (QDisc(..))
 import Data.Time.Units (Microsecond)
+
+blockingQDisc :: (EndPointAddress -> IO ()) -> IO (QDisc t)
+blockingQDisc waitForUnblock = do
+    fqd <- newFairQDisc
+    return QDisc {
+          qdiscEnqueue = \addr _ t -> do
+                  waitForUnblock addr
+                  writeFairQDisc fqd t
+        , qdiscDequeue = readFairQDisc fqd
+        }
 
 -- | Make a fair 'QDisc'. It's called fair, but it can also be unfair if you
 --   want it to be, by having some 'EndPointAddress's delay before writing.
--- a/src/Network/RateLimiting.hs
+++ b/src/Network/RateLimiting.hs
@@ -7,8 +7,8 @@ module Network.RateLimiting
          RateLimiting(..)
        , rlLift
          -- * Rate-limiting strategies
-       , rateLimitingUnbounded
-       , rateLimitingFair
+       , noRateLimitingUnbounded
+       , noRateLimitingFair
        , rateLimitingBlocking
        , rateLimitingDamping
        ) where
@@ -29,16 +29,28 @@ import           Network.Transport.TCP (QDisc(..), simpleUnboundedQDisc)
 -- This includes a 'QDisc' to govern how incoming requests are
 -- enqueued and dequeued, and also the possibility to block or delay
 -- requests from a peer depending on the number of in-flight bytes.
+--
+-- Each peer will get a "lock" that determines whether a new request
+-- will be immediately enqueued, delayed by a fixed time, or blocked
+-- until some requests from that peer are completed.
 data RateLimiting m =
       NoRateLimiting !(forall t. IO (QDisc t))
     | RateLimiting
-      { rlQDisc :: forall t. IO (QDisc t)
+      { -- | `QDisc` used for rate limiting.
+        rlQDisc :: forall t. IO (QDisc t)
+        -- | Change the state of the lock for a given peer, depending
+        -- on the number of in-flight bytes for this peer.  To be
+        -- called when the number of in-flight bytes changes.
       , rlLockByBytes
             :: (Mockable SharedAtomic m)
             => EndPointAddress
-            -> Int
+            -> Int -- ^ Number of in-flight bytes.
             -> m ()
+        -- | Register a lock for a new peer.  This will be called by a
+        -- node when a new peer connects
       , rlNewLock :: EndPointAddress -> m ()
+        -- | Unregister the lock for a peer.  To be called when the
+        -- last handler of a peer is finished.
       , rlRemoveLock :: EndPointAddress -> m ()
       }
 
@@ -53,13 +65,13 @@ rlLift RateLimiting{..} = RateLimiting
 
 -- | The simplest rate-limiting procedure uses an unbounded queue, and
 -- performs no rate limiting at all.
-rateLimitingUnbounded :: Monad m => RateLimiting m
-rateLimitingUnbounded = NoRateLimiting simpleUnboundedQDisc
+noRateLimitingUnbounded :: Monad m => RateLimiting m
+noRateLimitingUnbounded = NoRateLimiting simpleUnboundedQDisc
 
 -- | Ensures fariness in queueing requests from different peers, but
 -- does no rate-limiting.
-rateLimitingFair :: Monad m => RateLimiting m
-rateLimitingFair = NoRateLimiting (fairQDisc (const $ return Nothing))
+noRateLimitingFair :: Monad m => RateLimiting m
+noRateLimitingFair = NoRateLimiting (fairQDisc (const $ return Nothing))
 
 -- | A 'RateLimiting' that actually does rate-limiting.
 --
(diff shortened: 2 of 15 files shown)


kantp:
@avieth Actually, let me add some tests for the new code before reviewing this. I'll ping you when I have them.

kantp:
@avieth Ok, it's ready for a review now.

src/Network/RateLimiting.hs
```diff
@@ -0,0 +1,113 @@
...
+    | RateLimiting
+      { rlQDisc :: forall t. IO (QDisc t)
+      , rlLockByBytes
+            :: (Mockable SharedAtomic m)
+            => EndPointAddress
+            -> Int
+            -> m ()
+      , rlNewLock :: EndPointAddress -> m ()
+      , rlRemoveLock :: EndPointAddress -> m ()
+      }
```
avieth: Need comments for these other two fields as well `rlNewLock` `rlRemoveLock`. Are `rlLockByBytes` and `rlNewLock` supposed to be used together somehow?

src/Network/RateLimiting.hs
```diff
@@ -0,0 +1,113 @@
...
+
+-- | The simplest rate-limiting procedure uses an unbounded queue, and
+-- performs no rate limiting at all.
+rateLimitingUnbounded :: Monad m => RateLimiting m
+rateLimitingUnbounded = NoRateLimiting simpleUnboundedQDisc
+
+-- | Ensures fariness in queueing requests from different peers, but
+-- does no rate-limiting.
+rateLimitingFair :: Monad m => RateLimiting m
+rateLimitingFair = NoRateLimiting (fairQDisc (const $ return Nothing))
```
avieth: `rateLimitingUnbounded` and `rateLimitingFair` should be called `noRateLimitingUnbounded` and `noRateLimitingFair`.

src/Network/RateLimiting.hs
```diff
@@ -0,0 +1,113 @@
...
+-- given threshold, further messages from that peer will be blocked,
+-- until the number of in-flight bytes drops below the threshold
+-- again.
+rateLimitingBlocking
+    :: (Mockable SharedAtomic m)
+    => (forall t. m t -> IO t)
+    -> Int -- ^ Maximum in-flight bytes per peer.
+    -> m (RateLimiting m)
+rateLimitingBlocking liftIO maxBytesPerPeer =
+    rateLimitingDamping liftIO (const Nothing) (Just maxBytesPerPeer)
```
avieth: Not so important I guess but: the typical `liftIO :: forall t . IO t -> m t` so this name is a bit misleading. How about `lowerIO`?

src/Network/RateLimiting.hs
```diff
@@ -0,0 +1,113 @@
...
+                Nothing -> return Nothing
+                Just lock -> readSharedAtomic lock
+        , rlLockByBytes = \peer bytes -> do
+            lockMap <- readSharedAtomic locks
+            case Map.lookup peer lockMap of
+                Nothing -> return ()
+                Just mvar ->
+                    if isJust mThreshold && fromJust mThreshold < bytes
+                        then void (tryTakeSharedAtomic mvar)
+                        else void (tryPutSharedAtomic mvar (dampingFunction bytes))
```
avieth: Why the `try` variants? Shouldn't we be using `readSharedAtomic` and `swapSharedAtomic` (doesn't exist currently)?

avieth: Oh never mind, I see what's going on.

But where is the `dampingFunction` actually used? It computes some `Microsecond`, but it seems they're never used. The `QDisc` will wait until the `lock` is filled, but shouldn't it take the `Microsecond` and use it in a `threadDelay`?

src/Node.hs
```diff
@@ -281,11 +282,12 @@ node
        )
     => NT.Transport m
     -> StdGen
+    -> RL.RateLimiting m
     -> packing
     -> peerData
     -> (Node m -> NodeAction packing peerData m t)
     -> m t
-node transport prng packing peerData k = do
+node transport prng rateLimiting packing peerData k = do
```
avieth: The `RateLimiting m` contains a `forall t . IO (QDisc t)`, but the `Transport m` presumably has one of these as well. I suppose it's the programmer's responsibility to ensure that the transport is created using `rlQDisc`. I think we should make this more clear by having the `RateLimiting m` parameter come first, and then take a `(forall t. IO (QDisc t)) -> NT.Transport m` which will always be called with the `rlQDisc` of the `RateLimiting m`.

src/Network/QDisc/Fair.hs
```diff
@@ -15,6 +16,16 @@ import Network.Transport (EndPointAddress)
...
+blockingQDisc :: (EndPointAddress -> IO ()) -> IO (QDisc t)
+blockingQDisc waitForUnblock = do
+    fqd <- newFairQDisc
+    return QDisc {
+          qdiscEnqueue = \addr _ t -> do
+                  waitForUnblock addr
+                  writeFairQDisc fqd t
+        , qdiscDequeue = readFairQDisc fqd
+        }
+
```
avieth: Seems this definition is not used. Was that unintentional, or should we remove it?

kantp: Right, that's not needed and we should remove it.

src/Network/RateLimiting.hs
kantp: Yes, that's right.

src/Network/RateLimiting.hs
kantp: The `Microsecond` will be used in `fairQDisc`: we're using a `fairQDisc` where `getDelay` reads the number of `Microsecond`s to wait from an `MVar`. Depending on the `dampingFunction` and `mThreshold`, this `MVar` will either contain some delay, which will be used by `fairQDisc`, or it will be held, in which case `getDelay` will block.
So `rateLimitingDamping` is more general than `rateLimitingBlocking`: it can delay messages, but it can also block, if there are too many in-flight bytes.

src/Network/RateLimiting.hs
```diff
@@ -0,0 +1,113 @@
...
+-- enqueued and dequeued, and also the possibility to block or delay
+-- requests from a peer depending on the number of in-flight bytes.
+data RateLimiting m =
+      NoRateLimiting !(forall t. IO (QDisc t))
+    | RateLimiting
+      { rlQDisc :: forall t. IO (QDisc t)
+      , rlLockByBytes
+            :: (Mockable SharedAtomic m)
+            => EndPointAddress
+            -> Int
```
kantp: You're right, I'm adding some comments.

src/Network/RateLimiting.hs
kantp: Yes, they should be called upon receiving a connection from a new peer and when closing the last handler for a given peer. I'll add a comment explaining that.

kantp:
I tried the approach of having the `startNode` take a `Node m -> m (NT.Transport m)` parameter, but I still needed to pass some rules for when to take/release a lock, since that has to be triggered each time we receive data from a peer or finish a handler. So I decided to bundle up the `QDisc` with a `Map` of `MVar`s for locking/delaying, and a function for modifying those `MVar`s. That's what the `RateLimiting` is.

However, there are still two problems with that, that you have already identified:

- The interface right now can only adjust the rate depending on the number of in-flight bytes, which is rather inflexible. Having the connection statistics available would be preferable. I'd suggest moving the `Statistics` to a separate module, so that we can use it in `RateLimiting` and `Node.Internal` without introducing circular dependencies.

- Possible mismatch of `QDisc`s used by `Transport` and `RateLimiting` in `node`. As you proposed, I'll change `node` to take a `(forall t. IO (QDisc t)) -> NT.Transport m` instead of the `Transport`.

kantp:
@avieth I tried replacing the argument of type `NT.Transport m` with a `((forall t. IO (QDisc t)) -> m (NT.Transport m))`, so that `node` itself will create and close the transport in a `bracket`.

While it is nice to have the resource management inside the `node`, and to have a guarantee that the `QDisc`s match, that is problematic for the way the tests are performed at the moment. The tests exploit that the `Transport` used by a node can be accessed from without. So if we want to go ahead with this change, we\ll have to change the tests and make sure that no client code relies on accessing the `Transport` directly.

src/Network/RateLimiting.hs
avieth: Ok. I forgot that this change to `fairQDisc` reached master.

avieth:
For the `RateLimiting` examples which actually do rate limiting, the map of delays is global to the `QDisc`, so delays enforced by any particular node/`EndPoint` on one transport will affect all other nodes/`EndPoint`s on that transport. This probably won't come up in any real world situation, but it certainly could come up in testing and simulation.

kantp:
You're right, and this would be prevented if we did have the `Transport` created and closed within `node` (and the fact that doing this would require us to change the way the tests work probably means that the tests are affected).

avieth:
Seems to me this trouble is caused by the fact that a `QDisc` is created for a transport. If it were created for each `EndPoint` then I think having `Node m -> m (QDisc t)` would be possible. I'm really keen to have rate limiting be a particular case of this, rather than having special infrastructure just for it. That's to say, in my opinion we need a general way of making a `QDisc` local to an `EndPoint`/node.

If we had `startNode` create the whole transport, that would work, but as you mentioned it's a problem for the existing tests. And it's quite a problem in general, since creating a transport will probably involve a lot of resource acquisition, and it would make it impossible to use an in-memory transport for a simulation with more than one node! So that's a no go for sure.

And the in-memory transport is doubly relevant: it takes no `QDisc` (there's really no point) and so having the `QDisc` factor into the general network-transport `newEndPoint` API doesn't make sense.

To summarize my main issue, it's not exactly about the implementation/changes in this pull request, but about the infrastructure that this pull request is forced to use: it's not clear whether it's possible to get mutable state for a `QDisc` local to each instantiation (`EndPoint`). I need to think about this some more.

avieth:
I'm fairly certain we can't do the right thing without making a change to network-transport-tcp. My idea is to have the `TransportInternals` give a special variant of `newEndPoint` with this type:

```
QDisc t -> IO (Either (TransportError NewEndPointErrorCode) EndPoint)
```

and then `node`/`startNode` will not take a `Transport` but instead an acquirer/releaser for an `EndPoint` with the node itself in scope:

```
Node m -> (m (EndPoint m), EndPoint m -> m ())
```

Now to make a node with a rate-limiting `QDisc`, the TCP's internals must be used, but no other special infrastructure is needed: just derive the `QDisc t` from the `Node m`.

The original network-transport `newEndPoint` will of course still use the same `IO (QDisc t)` for all `EndPoint`s.

kantp:
Yes, that sounds good.

Should we proceed such that you implement the necessary changes in `network-transport-tcp`, while I add facilities for global rate-limiting?

kantp:
Or, we could go with implementing global rate-limiting first. @dcoutts suspects that we might not need a per-peer rate limiting anyway if we're queueing fairly and implement global rate-limiting.

avieth:
> Should we proceed such that you implement the necessary changes in network-transport-tcp, while I add facilities for global rate-limiting?

I think all necessary changes to time-warp-nt are here https://github.com/serokell/time-warp-nt/pull/35

network-transport-tcp changes are linked in that PR.

avieth:
Heads up: https://github.com/haskell-distributed/network-transport-tcp/pull/58 was merged to nt-tcp master and it should help with rate limiting.

```Haskell
-- TransportInternals gives us this:
newEndPointInternal :: (forall t . Maybe (QDisc t)) -> IO (Either (TransportError NewEndPointErrorCode) EndPoint)

-- A node requires this type as its first parameter
type MakeNewEndPoint peerData m = (SharedAtomicT m (LL.NodeState peerData m) -> LL.NodeEndPoint m)
```

So the `SharedAtomicT m (LL.NodeState peerData m)` can be in scope when the `QDisc` is defined, which should make things simpler.

PR #35, opened 2017-03-20
node/startNode don't take a Transport

avieth:
Instead, they just need a way to create and to close an EndPoint. For
the low-level startNode, the whole Node is in scope. For the high level
node, only the mutable cell of the node's state is in scope.

This allows us to use non-standard features of a transport, for instance
the TCP transport's QDisc, to create an EndPoint to be used by the node.
There is a patch for network-transport-tcp which allows the choice of a
particular QDisc for each new EndPoint, but it can only be done through
the TCP internals as the network-transport newEndPoint type doesn't
permit such a thing. With the changes here, it's possible to use the TCP
internals to create a special QDisc with the node's state in scope,
thereby using traffic metrics to control queueing.

Relevant network-transport-tcp change is [here](https://github.com/haskell-distributed/network-transport-tcp/pull/58)


kantp:
Looks good to me (though we should probably do the `TODO document` at `data NodeEndPoint` before merging).

PR #36, opened 2017-03-21
Fix typo in pstAddHandler

avieth:
This caused many warnings to appear in self-connection tests. A self
connection makes a local and a remote entry (it's both, technically) but
adding the remote handler wouldn't bump the count in the map, so it was
removed prematurely.


avieth:
Oops. Nothing to see here :) @kantp

PR #38, opened 2017-04-13
Use nt-tcp NAT, hoist functions for abstract transport

avieth:
In one application it was necessary to change the monad over which a transport operates. `hoistTransport` implements this.


avieth:
@georgeee This one is simple but important. Use of latest network-transport-tcp interface (with support for unaddressable nodes) ought to be in master, and is provided here. The `hoistTransport` is needed for the cardano-sl discovery abstraction branch.

avieth:
@georgeee Please approve of request changes.

PR #39, opened 2017-04-13
Correct dispatcher exception throwing / handling

avieth:
A fairly egregious oversight by me! The application needs to know if the
dispatcher stops doing its thing.

### DIFF
--- a/src/Node/Internal.hs
+++ b/src/Node/Internal.hs
@@ -814,7 +814,15 @@ nodeDispatcher node handlerIn handlerInOut =
             return (st, ())
 
         _ <- waitForRunningHandlers node
-        return ()
+
+        -- Check that this node was closed by a call to 'stopNode'. If it
+        -- wasn't, we throw an exception. This is important because the thread
+        -- which runs 'startNode' must *not* continue after the 'EndPoint' is
+        -- closed.
+        withSharedAtomic nstate $ \nodeState ->
+            if _nodeStateClosed nodeState
+            then pure ()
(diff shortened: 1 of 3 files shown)


src/Node/Internal.hs
```diff
@@ -603,6 +603,8 @@ startNode packingType peerData mkNodeEndPoint prng nodeEnv handlerIn handlerOut
                                 }
                       ; dispatcherThread <- async $
                             nodeDispatcher node handlerIn handlerOut
+                      -- Exceptions in the dispatcher are re-thrown here.
+                      ; link dispatcherThread
```
avieth: I would have preferred to use a `withAsync` here but as far as I can tell it's not feasible due to the mutual dependence of `dispatcherThread` and `node`.

src/Node/Internal.hs
```diff
@@ -1161,12 +1169,7 @@ nodeDispatcher node handlerIn handlerInOut =
 
                             Channel.writeChannel channel Nothing
                             return channels'
-                        (Nothing, channels') -> do
-                            logWarning $ sformat "inconsistent peer and connection identifier state"
-                            return channels'
-                        (Just cState, _channels') -> do
-                            logError $ sformat ("unexpected ConnectionState in connectionLost: " % shown) cState
-                            throw $ InternalError "nodeDispatcher: connectionLost: impossible"
```
avieth: For the record I have no idea why I ever wrote this. It's not impossible to lose a heavyweight connection while some lightweight connection is doing the handshake.

PR #41, opened 2017-04-19
NAT support with special case for 0.0.0.0, 127.0.0.1

avieth:
Emergency patch to handle NAT without forcing users to upgrade their clients.

The meat is in the network-transport-tcp revision bump avieth/csl_2_nat https://github.com/avieth/network-transport-tcp/tree/avieth/csl_2_nat

### DIFF
--- a/bench/Receiver/Main.hs
+++ b/bench/Receiver/Main.hs
@@ -40,7 +40,7 @@ main = do
     loadLogConfig logsPrefix logConfig
     setLocaleEncoding utf8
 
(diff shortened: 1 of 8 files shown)


avieth:
Superseded by #38 . It includes strictly more improvements. The only thing it's missing that this one has is the bump of the network-transport-tcp revision to the one with the hack for `0.0.0.0` and `127.0.0.1`, but so long as CSL always uses that revision it's ok, no?

avieth:
Closing. All changes in here can be found in other PRs. The network-transport-tcp revision is not important for this repo, but for CSL.

PR #44, opened 2017-04-22
[CSL-959] Add withSocketDo

Martoon-00:
And without `MonadBaseControl IO` unlike in previous commit :(

### DIFF
--- a/src/NTP/Util.hs
+++ b/src/NTP/Util.hs
@@ -7,19 +7,24 @@ module NTP.Util
     , preferIPv6
     , selectIPv6
     , selectIPv4
+    , withSocketsDoLifted
     ) where
 
-import           Control.Monad.Catch   (catchAll)
-import           Control.Monad.Trans   (MonadIO (..))
-import           Data.List             (sortOn)
-import           Data.List             (find)
-import           Data.Time.Clock.POSIX (getPOSIXTime)
-import           Data.Time.Units       (Microsecond, fromMicroseconds)
-import           Network.Socket        (AddrInfo, AddrInfoFlag (AI_ADDRCONFIG),
-                                        Family (AF_INET, AF_INET6), PortNumber (..),
-                                        SockAddr (..), SocketType (Datagram), addrAddress,
-                                        addrFamily, addrFlags, addrSocketType,
-                                        defaultHints, getAddrInfo)
+import           Control.Concurrent          (forkIO)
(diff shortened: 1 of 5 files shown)


src/NTP/Util.hs
```diff
@@ -64,3 +69,14 @@ selectIPv6 = find (\a -> addrFamily a == AF_INET6)
...
 selectIPv4 = find (\a -> addrFamily a == AF_INET)
+
+withSocketsDoLifted :: (MonadIO m, MonadMask m) => m a -> m a
+withSocketsDoLifted action = do
+    -- `MonadBaseControl IO` is not provided sometimes, so hack with threads
+    -- is used
+    bracket (liftIO $ newTVarIO False)
+            (liftIO . atomically . flip writeTVar True) $
+            \exited -> do
+                liftIO . void . forkIO $
```
ChShersh: Why this should be done in separate fork?

Martoon-00: `withSocketsDo` has signature `IO a -> IO a`, as argument I have to provide an action, which works till I need networking capabilities.
This it the only way I see, without using `liftBaseWith`

ChShersh: Could you put these thougs into comments before this function then? Even better with Haddock. After doing that I can approve and merge.

Martoon-00: Why haddock? It's implementation detail

ChShersh: Well, it affects behaviour, because implementation creates new thread. I think this can be relevant, because you're not just adding type constraints and not only adding `liftIO`.
Haddock is just good doc, why not write more docs? But at least I want such thoughts to be written so somebody can look at this later and understand, what's going on.

Martoon-00:
Comment to `withSocketsDoLifted` added

src/NTP/Util.hs
int-index: @Martoon-00 Why do you think this solution is nicer? I'd avoid forking whenever possible.

int-index: To clarify why forking is bad - it obscures control flow and makes debugging a nightmare.

Martoon-00:
Okay, I overestimated number of missing `MonadBaseControl IO` instances in cardano, let's close this issue as soon as https://github.com/input-output-hk/cardano-sl/pull/281 gets merged

PR #48, opened 2017-04-24
Merge `avieth/nat` to `time-warp-nt-0.1.1`

Martoon-00:
This is required for `cardano-sl:cardano-sl-0.4` to `cardano-sl:master` merge

### DIFF
--- a/src/Network/Discovery/Transport/Kademlia.hs
+++ b/src/Network/Discovery/Transport/Kademlia.hs
@@ -73,7 +73,7 @@ kademliaDiscovery configuration initialPeer myAddress = do
         port = fromIntegral (kademliaPort configuration)
     -- A Kademlia instance to do the DHT magic.
     kademliaInst :: K.KademliaInstance (KSerialize i) (KSerialize EndPointAddress)
-        <- liftIO $ K.create "127.0.0.1" port kid
+        <- liftIO $ K.create (kademliaHost configuration) port kid
     -- A TVar to cache the set of known peers at the last use of 'discoverPeers'
     peersTVar :: TVar.TVar (M.Map (K.Node (KSerialize i)) EndPointAddress)
         <- liftIO . TVar.newTVarIO $ M.empty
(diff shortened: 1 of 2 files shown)


src/Network/Discovery/Transport/Kademlia.hs
```diff
@@ -73,7 +73,7 @@ kademliaDiscovery configuration initialPeer myAddress = do
         port = fromIntegral (kademliaPort configuration)
     -- A Kademlia instance to do the DHT magic.
     kademliaInst :: K.KademliaInstance (KSerialize i) (KSerialize EndPointAddress)
-        <- liftIO $ K.create (kademliaHost configuration) port kid
+        <- liftIO $ K.create "127.0.0.1" port kid
```
avieth: This line shouldn't be changed. The old version is correct.

stack.yaml
```diff
@@ -3,16 +3,16 @@ resolver: lts-8.2
...
+      commit: 21df94f41008f82ee023e8e0324c7e4a82c4fef2
     extra-dep: true
   - location:
       git: https://github.com/avieth/network-transport
       commit: f2321a103f53f51d36c99383132e3ffa3ef1c401
     extra-dep: true
   - location:
       git: https://github.com/avieth/network-transport-tcp
-      commit: ca42a954a15792f5ea8dc203e56fac8175b99c33
+      commit: eb1ed2fe4d4419c860ce060cb1091f7c8959134f
```
avieth: Just a note: the old revision comes immediately after the new revision in `avieth/csl_2_nat` branch of network-transport-tcp. Cardano-sl needs to be built with `ca42a954` because it includes a (temporary) emergency patch which makes wallet/daedalus appear as unaddressable. It should be OK to use `eb1ed2fe` here though.

Martoon-00: Old commit is used in cardano as well, I would better return it

Martoon-00:
@avieth Is it ready?

PR #50, opened 2017-04-24
[CSL-1075] Kademlia configurable bind host

avieth:
An old branch that ought to be merged.


georgeee:
P.S. Before merging tag with "CSL-1075 Kademlia: make bind host configurable"

PR #51, opened 2017-04-24
Fix benchmark sender import, tests, for nt-tcp interface change

avieth:
network-transport-tcp revision bump caused benchmark sender to fail because of an export change.


avieth:
This fixes CI failures in other PRs. Looking for some approval @georgeee

PR #52, opened 2017-04-25
Fix compilation on nix

volhovM:
cc @Martoon-00 @ChShersh 
this makes `stack --nix build` happen.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -20,8 +20,7 @@ packages:
     extra-dep: true
 
 nix:
-  enable: false
-  packages: []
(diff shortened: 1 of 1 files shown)


ChShersh:
@volhovM Btw, do you know: is it enough to have `nix-shell` on _Ubuntu_ to use automatically use it with `stack --nix`?

volhovM:
@ChShersh I suppose so, yes. What `stack --nix` does internally is just runs your `stack --no-nix` command in `nix-shell` environment.

PR #53, opened 2017-04-26
Bump cabal version to 0.1.2

Martoon-00:


Martoon-00:
Perhaps better to rename `time-warp-nt-0.1.1` into `time-warp-nt-0.1`

PR #54, opened 2017-04-27
[TW-138] provide peer data on all interactions

avieth:
Previously, sending a single message would not show the peer data. Now it does, because a single message interaction is just a conversation where one of the send or receive types is `Void`. The conversation does not proceed until the peer data of the remote peer is available, as it always will be because they must ACK the request, and peer data is always available if there's at least one incoming connection. (There is a longstanding but rare bug caused by `EventConnectionLost` which I'll fix soon).

Also included: the set of listeners is computed from the peer data each time a conversation is begun. In a typical application where the peer data specifies version information, this could be very useful: the set of listeners can depend upon the version info, rather than having a fixed set of listeners, each of which depends upon the version info (some of them probably wouldn't make any sense for certain versions).

This change may (probably will) negatively affect performance. Previously, some interactions could go ahead without waiting for the peer data. Now, necessarily, they will all wait for it. There's no avoiding that, but ideally the peer data would remain in memory across multiple interactions with a given peer. A simple way to improve things is to have a keepalive connection which stays up for, say, 30 seconds after the last application connection goes down. We could do that, but I think it's the wrong way around...

After many days of considering other options, I went with this solution because we need something now, and everything else is too radical and would require too much time. But I want to talk about the right way around so that we may eventually come to an agreement and schedule the work to be done.

The motivation for some keepalive connection comes from how time-warp is used in cardano-sl: threads contact peers basically whenever and wherever, with the understanding that some magic in time-warp will make everything maximally efficient. Since there's no application-level structure to the interaction with a given peer, all time-warp can do is guess that you may contact a peer again some time soon, hence the keepalive. I believe explicit sessions as described in #24 are a better solution. No keepalive necessary: the application explicitly delimits an interaction with a peer, in which the peer data is available and during which there need not be any conversations taking place. Instead of contacting peers whenever and wherever, an application is encouraged to establish sessions with those peers with which it anticipates recurring interaction. Those interactions are localized to the thread which establishes the session, and the nature of the interactions itself can depend upon the peer data, which is held across multiple conversations rather than delivered anew for each one.

An example to make things a little more concrete: instead of periodically broadcasting a piece of data to all neighbours by starting a bunch of independent conversations (which may require exchanging peer data), have a long-running *explicit* session with each of these neighbours, a thread for each, and push the data to broadcast to these threads.

tl;dr time-warp should not attempt to compensate for deficiencies in the architecture of the application (cardano-sl)

### DIFF
--- a/bench/Receiver/Main.hs
+++ b/bench/Receiver/Main.hs
@@ -22,7 +22,7 @@ import           Bench.Network.Commons      (MeasureEvent (..), Ping (..), Pong 
 import qualified Network.Transport.TCP      as TCP
 import           Network.Transport.Concrete (concrete)
 import           Node                       (ListenerAction (..), NodeAction (..), node,
-                                             sendTo, defaultNodeEnvironment,
+                                             defaultNodeEnvironment, ConversationActions (..),
                                              simpleNodeEndPoint)
 import           Node.Message               (BinaryP (..))
 import           ReceiverOptions            (Args (..), argsParser)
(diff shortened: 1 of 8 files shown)


georgeee:
TODO before merge: create issue to consider Alex's proposal on explicit sessions and keep-alive later

ChShersh:
Am I correct that this PR completely removes _one-message style_ exhanging? And all code in `cardano-sl` should be refactored in appropriate way (though this refactoring makes code simpler).

But can some optimizations be made? For example, we can introduce function like `recvVoid` which will send `CONNECTION CLOSED` after fully receiving and parsing data. So this function can be used instead of old one-message handlers. 

Maybe sending `CONNECTION CLOSED` after receiving data is better for performance if old connections in this new scheme are not reused for single style messaging (or this should go to _Application Level_ how we handle connection reusage?).

PR #55, opened 2017-04-28
[TW-144] ignore incoming data if handler finished

avieth:
This prevents a simple attack (described in the ticket) in which anybody
who knows what peer data you expect can fill up your heap with bytes.

Note that there's another side to the story in case this is backed by
network-transport-tcp: the transport's QDisc. If the simple unbounded
QDisc is used, the attack will probably still work, because the
unbounded Chan will fill up with data faster than the dispatcher can
remove it. Using a one-place or fair QDisc mitigates this.

### DIFF
--- a/src/Node/Internal.hs
+++ b/src/Node/Internal.hs
@@ -40,7 +40,6 @@ module Node.Internal (
     closeChannel,
     startNode,
     stopNode,
-    withOutChannel,
     withInOutChannel,
     writeMany,
     Timeout(..)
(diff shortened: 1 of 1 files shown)


avieth:
@georgeee @dcoutts if we're concerned about DoS then we should merge this one.

src/Node/Internal.hs
```diff
@@ -1056,6 +1059,7 @@ nodeDispatcher node handlerInOut =
                           -- No matter what, we must update the node state to
                           -- indicate that we've disconnected from the peer.
                           let cleanup conn (me :: Maybe SomeException) = do
+                                  modifySharedAtomic chanVar $ \_ -> return (Nothing, ())
```
avieth: @dcoutts right here

avieth:
This prevents an attack by a peer who SYNs the conversation, but looking through the changes just now I realize the same attack is still possible by an ACKing peer. Will fix that.

src/Node/Internal.hs
dcoutts: Thanks

PR #58, opened 2017-05-02
[TW-145] node can delay receives

avieth:
A 'ReceiveDelay' value can be given, defined possibly in terms of the
node state. It will be run before each network-transport 'receive' and
if a 'Just' is given, the dispatcher thread will delay for that many
microseconds before calling receive. The idea is that a sufficiently
smart choice for this can help a node deal with excess load.


avieth:
This snuck into master with #59

PR #59, opened 2017-05-09
[DEVOPS-74] Setup monitoring without starting server

avieth:
This will be useful in application which wish to extend the EKG server with their own metrics.


avieth:
Will need this for DEVOPS-74 so please review @kantp @domenkozar

domenkozar:
@avieth if I understand correctly there will be a seprate PR to integrate https://hackage.haskell.org/package/ekg-statsd?

avieth:
> @avieth if I understand correctly there will be a seprate PR to integrate https://hackage.haskell.org/package/ekg-statsd?

Yes, but it will need this change. I could make a new branch with it but it'd be simpler to bring it into master first.

PR #60, opened 2017-05-10
Log our timewarp NodeId when we sart up a node

dcoutts:
This will help in our logs to identify nodes.


dcoutts-2fa:
@georgeee could you check and merge?

PR #61, opened 2017-05-29
Georgeee/csl 831 allow dynamic message limit for listeners

georgeee:
PR contains:
1. Merge of `master`
2. Use of latest kademlia
3. Change in interface (listeners are now retrieved as monadic computation)

### DIFF
--- a/src/Node.hs
+++ b/src/Node.hs
@@ -158,7 +158,7 @@ hoistSendActions nat rnat SendActions {..} = SendActions withConnectionTo'
         Conversation l -> Conversation $ \cactions -> rnat (l (hoistConversationActions nat cactions))
 
 type ListenerIndex packing peerData m =
-    Map MessageName (ListenerAction packing peerData m)
+    Map MessageName (Listener packing peerData m)
 
 makeListenerIndex :: [Listener packing peerData m]
                   -> (ListenerIndex packing peerData m, [MessageName])
@@ -233,7 +233,7 @@ nodeConversationActions _ _ packing inchan outchan =
             Input t -> pure (Just t)
 
 data NodeAction packing peerData m t =
-    NodeAction (peerData -> [Listener packing peerData m])
+    NodeAction (peerData -> m [Listener packing peerData m])
                (SendActions packing peerData m -> m t)
 
 simpleNodeEndPoint
@@ -285,8 +285,8 @@ node mkEndPoint prng packing peerData nodeEnv k = do
           -- Index the listeners by message name, for faster lookup.
           -- TODO: report conflicting names, or statically eliminate them using
           -- DataKinds and TypeFamilies.
-        ; let listenerIndices :: peerData -> ListenerIndex packing peerData m
-              listenerIndices = fmap (fst . makeListenerIndex) mkListeners
+        ; let listenerIndices :: peerData -> m (ListenerIndex packing peerData m)
+              listenerIndices = fmap (fst . makeListenerIndex) <$> mkListeners
         ; llnode <- LL.startNode
               packing
               peerData
@@ -320,14 +320,14 @@ node mkEndPoint prng packing peerData nodeEnv k = do
     -- message name, then choose a listener and fork a thread to run it.
     handlerInOut
         :: LL.Node packing peerData m
-        -> (peerData -> ListenerIndex packing peerData m)
+        -> (peerData -> m (ListenerIndex packing peerData m))
         -> peerData
         -> LL.NodeId
         -> ChannelIn m
         -> ChannelOut m
         -> m ()
     handlerInOut nodeUnit listenerIndices peerData peerId inchan outchan = do
-        let listenerIndex = listenerIndices peerData
+        listenerIndex <- listenerIndices peerData
         input <- recvNext inchan packing
         case input of
             End -> logDebug "handlerInOut : unexpected end of input"
--- a/src/Node.hs
+++ b/src/Node.hs
@@ -19,6 +19,9 @@ module Node (
       Node(..)
     , LL.NodeEnvironment(..)
     , LL.defaultNodeEnvironment
+    , LL.ReceiveDelay
+    , LL.noReceiveDelay
+    , LL.constantReceiveDelay
     , nodeEndPointAddress
     , NodeAction(..)
     , node
@@ -238,7 +241,7 @@ data NodeAction packing peerData m t =
 
 simpleNodeEndPoint
     :: NT.Transport m
-    -> SharedAtomicT m (LL.NodeState peerData m)
+    -> m (LL.Statistics m)
     -> LL.NodeEndPoint m
 simpleNodeEndPoint transport _ = LL.NodeEndPoint {
       newNodeEndPoint = NT.newEndPoint transport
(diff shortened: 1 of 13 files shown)


src/Node.hs
```diff
@@ -320,14 +325,14 @@ node mkEndPoint prng packing peerData nodeEnv k = do
     -- message name, then choose a listener and fork a thread to run it.
     handlerInOut
         :: LL.Node packing peerData m
-        -> (peerData -> ListenerIndex packing peerData m)
+        -> (peerData -> m (ListenerIndex packing peerData m))
```
georgeee: @avieth this is major change within this PR

avieth:
I remember discussing this many weeks ago. Was there some interim solution?

I'm still sceptical. By giving the listeners up-front, without the `m` context, we know that they're static, stateless, time-independent. We would need a very good reason to give that up and I'm not sure we have one. Will it not suffice to paginate input? Do many `recv`s of messages of bounded size, and close the connection if too many messages come in.

georgeee:
@avieth 
Concern is same as some weeks ago. We need to refine limitations on message size depending on monadic context. There is no way we can do it differently, because these limits *are* dynamic and computed based on state of blockchain.
And few weeks ago I explained to you, why we need exactly this kind of limitation
Only way you proposed to not expose this monadic context is to move limit logic to time-warp. But this would lead to:
* Effort to move code
* Time-warp would represent logic that is higher level

This is doable, but:
* We have no time for this
* Solution with listeners returned in monad does effectively same thing
* We need solution now, because without it our update system would for sure be broken

avieth:
> There is no way we can do it differently, because these limits are dynamic and computed based on state of blockchain. And few weeks ago I explained to you, why we need exactly this kind of limitation

I agree and I understand why/how this comes about. I just think we're dealing with it in the wrong way. One particularly bad consequence is that we have the typeclasses `MessageLimited` and `MessageLimitedPure`. They give magic numbers which are implicit in some serialize/deserialize instance. If the shape of the datatype, or just its encoding in bytes, should ever change, the developer must know to update these otherwise completely unrelated instances (and also correctly compute the new limit, which may not be obvious).

The limit, if it exists, is implicit in the deserialize instance; there's no need to re-state it. The real problem is that we're choosing input types which have no such limit. For listeners which must accept such input, it should be reconstructed from smaller chunks according to the state of the blockchain, which can already be accessed in the body and used to control `recv`s and `send`s.

> Only way you proposed to not expose this monadic context is to move limit logic to time-warp. But this would lead to:
>
> - Effort to move code
> - Time-warp would represent logic that is higher level

My proposal is to use input types such that their deserializers will fail when too much input is given. time-warp would say nothing of message size limits. It would require no changes, but cardano-sl would. The listeners are implemented in the application's monad and so they can get the data they need in order to decide when to stop receiving more input.

> - We have no time for this
> ...
> - We need solution now, because without it our update system would for sure be broken

georgeee:
> The real problem is that we're choosing input types which have no such limit.
> My proposal is to use input types such that their deserializers will fail when too much input is given

Sorry, but this sound to me as some unicornish stuff.
How exactly are you going to limit block? Which at some point may contain payload of up to X mb of transactions and stuff, at another point of up to Y?
We discussed already your idea to decompose block type and send separately every part of it. And have limit on transaction amount being checked on app level.
And then somehow limit size of transaction by (probably) limitting number of inputs + outputs + txDistr + something else.
And we will come up with solution which would be:
* way more complex (more limits)
* harder to use in practice (we would need to compute estimation on size of that or other part depending on these limits instead of just setting strict limit for bigger structure)
* in principle harder to handle on `app` level (instead of sending single data we will have to write some ad-hoc logic to send/receive this data in parts, counting it, and not only for blocks -- transactions, update proposals, ssc payloads as well)

And this solution would imply huge refactoring of CSL.
Again, I remember I brought these points some months ago.

georgeee:
> I agree and I understand why/how this comes about. I just think we're dealing with it in the wrong way. One particularly bad consequence is that we have the typeclasses MessageLimited and MessageLimitedPure. They give magic numbers which are implicit in some serialize/deserialize instance. If the shape of the datatype, or just its encoding in bytes, should ever change, the developer must know to update these otherwise completely unrelated instances (and also correctly compute the new limit, which may not be obvious).
BTW, it's not exatcly true. Most types which consists of several fields are limited in natural way, just similar to binary instances

avieth:
> How exactly are you going to limit block? Which at some point may contain payload of up to X mb of transactions and stuff, at another point of up to Y?

By receiving the data in chunks of bounded size. Here's a naive sketch of how it might work.

```Haskell
recvChunks limit currentSize chunks = do
    | currentSize > limit = return TooManyChunks
    | otherwise = do
        -- Brings in at most, say, 4096 bytes.
        -- If extra bytes are sent, the parse fails, and we'll stop accepting more.
        nextChunk <- recvNextChunk
        case nextChunk of
            Nothing -> return ChunkTooBig
            Just NoMore -> return $ Block (concatBlockChunks (reverse chunks))
            Just (More chunk) -> recvChunks limit (currentSize + size chunk) (chunk : chunks)
```

Rather than limiting the size of a block (which is somewhat artificial as you say because they don't have a fixed limit) we limit the size of the chunk used to reconstruct the block. The limit is just a wire-protocol thing, it's not a property of the crypto/blockchain datatypes themselves, as it *appears* to be in the current implementation.

> We discussed already your idea to decompose block type and send separately every part of it. And have limit on transaction amount being checked on app level.
> And then somehow limit size of transaction by (probably) limitting number of inputs + outputs + txDistr + something else.

Sure, we would limit the size of a transaction using the same formula we do now, namely `DB.gsMaxTxSize`. But instead of trying to attach this limit to the types `TxAux`, `TxMsgContents`, we would receive them in chunks and limit the size of those chunks.

> way more complex (more limits)

Right we introduce some limits on the size of chunks. That's a negligible increase in complexity considering the fact that we already have so many limits. On the contrary, my proposal will make it far *simpler* by elimination of these `MessageLimited` and `MessageLimitedPure` classes, and a static, stateless, time-independent listener set.

> harder to use in practice (we would need to compute estimation on size of that or other part depending on these limits instead of just setting strict limit for bigger structure)

I don't understand. We already have the limits: `DB.gsMaxTxSize`, `DB.gsMaxProposalSize`, `DB.gsMaxBlockSize`, `DB.gsMaxHeaderSize`, etc. These would not change. With reference to the example I posted above, these would serve as the `limit` arguments, and we'd compute them inside the listener body.

> in principle harder to handle on app level (instead of sending single data we will have to write some ad-hoc logic to send/receive this data in parts, counting it, and not only for blocks -- transactions, update proposals, ssc payloads as well)

This is a fairly simple abstraction. Ah, I suppose this *could* require a time-warp change, namely to implement the pagination/chunking pattern.

> And this solution would imply huge refactoring of CSL.

Admitted, but I hope to convince you that it's worth the effort.

avieth:
> > harder to use in practice (we would need to compute estimation on size of that or other part depending on these limits instead of just setting strict limit for bigger structure)
>
> I don't understand. We already have the limits: DB.gsMaxTxSize, DB.gsMaxProposalSize, DB.gsMaxBlockSize, DB.gsMaxHeaderSize, etc. These would not change. With reference to the example I posted above, these would serve as the limit arguments, and we'd compute them inside the listener body.

I think I see what you're getting at now. Many of the types we deserialize from the wire have an implicit limit on their binary encoding. A `PVSS.Commitment`, for example, [will take 33 bytes and leave the rest](https://hackage.haskell.org/package/pvss-0.1/docs/src/Crypto-PVSS-ECC.html#line-103). If somebody tries to give more, they won't succeed in filling up our heap with junk. Either we try to parse something else later and the junk bytes make it fail, or we're finished and the junk bytes are collected.

So it's not necessary to estimate the limit on the bigger structure (this is, by the way, exactly what we do now, and in many cases it really is just an estimate--not a tight bound, but a "should work" bound). Instead, we deserialize only types which are already bounded by their binary instances, and then construct the composite structure from them.

PR #62, opened 2017-06-01
[TW-148] import relay stuff from CSL

avieth:
Inv/Req/Data relaying mechanism spliced from cardano-sl and stripped down to what (I'm fairly sure) is minimally sufficient for relay broadcast.

```Haskell
-- 'relayer' relays a message, in this particular case by putting it into a queue
-- which is cleared by 'relayWorker', in the same way that the current cardano-sl
-- implementation does it (concurrent send to all neighbours).
-- However this one is parameterized by the way in which the targets are
-- determined.
(relayer, relayWorker) <- simpleRelayer (determineNextTarget :: Maybe NodeId -> m (Set NodeId))
-- Relay listeners need to know how to relay (a field in the 'Relay' datatype).
let listeners = relayListeners
        [ relayGroup1 relayer :: Relay packingType m
        , relayGroup2 relayer :: Relay packingType m
        ]
-- An example 'NodeAction': it has the relay listeners and immediately spawns a
-- thread to execute the relays in the usual naive single-queue way.
let nodeAction :: NodeAction packingType peerData m
    nodeAction = NodeAction (const listeners) $ \sactions ->
        wtihAsync (relayWorker sactions) $ \relayThread -> do
            -- You can initiate a relay like this.
            -- The 'Nothing' has type 'Maybe NodeId', giving the provenance of the
            -- relay message ('Just peer' means we're relaying for that peer).
            relayer (InvReqDataPM Nothing someKey someValue)
            ...
        
...
```

### DIFF
--- a/src/Network/Broadcast/Relay/Logic.hs
+++ b/src/Network/Broadcast/Relay/Logic.hs
@@ -1,0 +1,322 @@
+-- | Framework for Inv\/Req\/Data message handling
+
+{-# LANGUAGE Rank2Types          #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies        #-}
+{-# LANGUAGE FlexibleContexts    #-}
+
+module Network.Broadcast.Relay.Logic
+    ( relayListeners
+    , simpleRelayer
+
+      -- | Listeners for Inv, Req, and Data
+    , handleInvL
+    , handleReqL
+    , handleDataL
+    ) where
+
+import           Universum
+import           Formatting                         (build, sformat, (%))
+import           System.Wlog                        (WithLogger, logDebug, logWarning)
+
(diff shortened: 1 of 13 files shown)


src/Network/Broadcast/Relay/Logic.hs
```diff
@@ -0,0 +1,322 @@
...
+    -> [Listener packingType peerData m]
+relayListeners = (>>= listenersForRelay)
+
+-- | Create a relayed which uses a single bounded queue for every message to
+--   be relayed. The first component is the relayed to be included in the
+--   Relay descriptor definitions (for particular data types) and the second
+--   component will, given network capabilities, forever pull from the queue and
+--   relay the messages (you should probably spawn a thread for it).
+--
+--   TODO use a bounded queue!
```
dcoutts: Yes. I presume this needs a mockable extension.

PR #63, opened 2017-06-05
[WIP] CSL-1122 introduce store

georgeee:


avieth:
How about something like this:

```Haskell
type Length = Word32

class Unpack packingType t where
    unpack :: packingType -> Length -> Decoder t

instance Binary t => Unpack BinaryP t where
    unpack _ = ...

instance Store t => Unpack StoreP t where
    unpack _ = ...
```

Messages will always be length-prefixed. We'll use it to run `unpack`. To support `store`, `unpack` can read the bytes into some buffer of the given length and then run `Store.decode`. The instances themselves can also immediately fail if it's determined that the length is too long (or even too short).

To impose limits, simply do not accept types of potentially unbounded size as input.

avieth:
Since #64 has been merged, `store` can be used for serialization (see `src/Node/Message/Store.hs`).

PR #64, opened 2017-06-09
[TW-149] Redesign of message serialization

avieth:
- Outgoing messages are split into chunks of at most some number of
  bytes. This value must be properly configured when running over
  network-transport-tcp, to be smaller than the maximum receive size.
- Node.Message has been split up and the Packable/Unpackable classes
  rolled into one class Serializable, which demands an incremental
  decoder which mimics the on from Data.Binary.Get
- In a conversation, 'recv' takes a limit on the number of bytes which
  will be read.
- Can use store instead of binary for deserialization, as shown in the
  implementation of Node.Message.Store

TODO: give a CBOR-backed implementation and use criterion to get some real
performance numbers.

### DIFF
--- a/src/Node.hs
+++ b/src/Node.hs
@@ -351,7 +351,7 @@ node mkEndPoint mkReceiveDelay prng packing peerData nodeEnv k = do
         -> m ()
     handlerInOut nodeUnit listenerIndices peerData peerId inchan outchan = do
         let listenerIndex = listenerIndices peerData
-        input <- recvNext packing messageSizeLimit inchan
+        input <- recvNext packing messageNameSizeLimit inchan
         case input of
             End -> logDebug "handlerInOut : unexpected end of input"
             Input msgName -> do
@@ -363,8 +363,8 @@ node mkEndPoint mkReceiveDelay prng packing peerData nodeEnv k = do
                     Nothing -> error ("handlerInOut : no listener for " ++ show msgName)
     -- Arbitrary limit on the message size...
     -- TODO make it configurable I guess.
-    messageSizeLimit :: Int
-    messageSizeLimit = 256
+    messageNameSizeLimit :: Int
+    messageNameSizeLimit = 256
 
 -- | Try to receive and parse the next message, subject to a limit on the
 --   number of bytes which will be read.
--- a/src/Node.hs
+++ b/src/Node.hs
@@ -91,13 +91,6 @@ data Node m = Node {
 nodeEndPointAddress :: Node m -> NT.EndPointAddress
 nodeEndPointAddress (Node addr _ _) = LL.nodeEndPointAddress addr
 
--- | Maximum transmission unit: you can't send more than this many bytes in
---   one call. It's assumed to be lower than the network-transport 
---
---   TODO make it configurable.
-mtu :: Word32
-mtu = 4096
-
 data Input t = Input t | End
 
 data LimitExceeded = LimitExceeded
@@ -214,6 +207,8 @@ nodeSendActions nodeUnit packing =
     SendActions nodeWithConnectionTo
   where
 
+    mtu = LL.nodeMtu (LL.nodeEnvironment nodeUnit)
+
     nodeWithConnectionTo
         :: forall t .
            LL.NodeId
@@ -242,9 +237,11 @@ nodeConversationActions
     -> ChannelIn m
     -> ChannelOut m
     -> ConversationActions snd rcv m
-nodeConversationActions _ _ packing inchan outchan =
+nodeConversationActions node _ packing inchan outchan =
     ConversationActions nodeSend nodeRecv
     where
+
+    mtu = LL.nodeMtu (LL.nodeEnvironment node)
 
     nodeSend = \body -> do
         LL.writeMany mtu outchan (packMsg packing body)
--- a/src/Node.hs
+++ b/src/Node.hs
@@ -13,6 +13,7 @@
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE StandaloneDeriving         #-}
 {-# LANGUAGE TypeApplications           #-}
+{-# LANGUAGE BangPatterns               #-}
 
 module Node (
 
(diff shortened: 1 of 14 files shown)


src/Node.hs
```diff
@@ -134,7 +154,7 @@ data ConversationActions body rcv m = ConversationActions {
 
        -- | Receive a message within the context of this conversation.
        --   'Nothing' means end of input (peer ended conversation).
-     , recv :: m (Maybe rcv)
+     , recv :: Word32 -> m (Maybe rcv)
```
dcoutts-2fa: Update docs here to say what the Wod32 is.

src/Node.hs
```diff
@@ -333,41 +351,58 @@ node mkEndPoint mkReceiveDelay prng packing peerData nodeEnv k = do
...
                 let listener = M.lookup msgName listenerIndex
                 case listener of
                     Just (ListenerActionConversation action) ->
                         let cactions = nodeConversationActions nodeUnit peerId packing inchan outchan
                         in  action peerData peerId cactions
                     Nothing -> error ("handlerInOut : no listener for " ++ show msgName)
+    -- Arbitrary limit on the message size...
+    -- TODO make it configurable I guess.
+    messageSizeLimit :: Int
+    messageSizeLimit = 256
```
dcoutts-2fa: So what is this? The limit on the conversation name/header?

src/Node/Internal.hs
```diff
@@ -237,20 +237,26 @@ newtype ChannelOut m = ChannelOut (NT.Connection m)
...
+    -> LBS.ByteString
     -> m ()
-writeChannel (ChannelOut conn) chunks = do
-    res <- NT.send conn chunks
-    -- TBD error handling? Throw exception or report it as a Left?
-    case res of
-      Left err -> throw err
-      Right _  -> pure ()
+writeMany mtu (ChannelOut conn) bss =
+    NT.send conn (chop bss >>= LBS.toChunks) >>= either throw pure
```
dcoutts-2fa: This isn't right. The NT.send is a vectored send, meaning it assembles a single data frame from multiple buffers. We want multiple data frames (each of a limited size). So we need to do a `mapM_ NT.send` instead, not a single `NT.send`.

src/Node/Message/Store.hs
```diff
@@ -0,0 +1,59 @@
...
+    unpackMsg _ = storeDecoder BS.empty
+
+storeDecoder :: Store.Store t => BS.ByteString -> Decoder t
+storeDecoder bs = Partial $ \mbs -> case mbs of
+    Nothing -> Fail BS.empty (fromIntegral (BS.length bs)) "Unexpected end of input (length prefix)"
+    Just bs' ->
+        let (front, back) = BS.splitAt 4 (BS.append bs bs')
+        in  if BS.length front == 4
+            then storeDecoderBody (NT.decodeWord32 front) BS.empty (Just back)
+            else storeDecoder front
```
dcoutts-2fa: I don't grok the else case here. What's going on?

src/Node.hs
avieth: Right, should be `messageNameSizeLimit`.

src/Node/Internal.hs
avieth: Oops, fix on the way.

src/Node/Message/Store.hs
avieth: The whole observed input so far `BS.append bs bs'` is split at the 4th byte. If the front isn't of length 4, then it must be shorter than 4 bytes, so we need to wait for more input. In this case, `back` is empty so we can forget about it.

src/Node/Internal.hs
```diff
@@ -239,14 +239,18 @@ closeChannel (ChannelOut conn) = NT.close conn
...
     -> LBS.ByteString
     -> m ()
-writeMany mtu (ChannelOut conn) bss =
-    NT.send conn (chop bss >>= LBS.toChunks) >>= either throw pure
+writeMany mtu (ChannelOut conn) bss = mapM_ sendUnit units
   where
+    sendUnit :: [BS.ByteString] -> m ()
+    sendUnit unit = NT.send conn unit >>= either throw pure
+    units :: [[BS.ByteString]]
+    units = fmap LBS.toChunks (chop bss)
```
dcoutts-2fa: Yes, nice, so we'll always get frames of the max mtu size (except last) irrespective of how the lazy bytestring chunks are formed.

src/Node/Message/Store.hs
dcoutts-2fa: Ahh, ok, we're waiting for more input. Yes ok.

avieth:
Some new commits today:

- Updated the tests to make use of the MTU for network-transport-tcp.
- Respect the MTU while sending peer data.
- Make the MTU configurable in the `NodeEnvironment`.

avieth:
I think I've managed to make CI reliably pass. I suppose these CI machines aren't granted many resources, because the test suite was just timing out, but it always ran quickly on my modest laptop.

I found one big improvement in enabling TCP_NODELAY. It makes sense that not having this flag set, and introducing a small MTU, would make the "heavy" payload tests (at most 99k long) grind to a halt. Every transmission would be small enough that the system would wait for an ACK before sending.

avieth:
@georgeee Can @dcoutts and I have write access?

Martoon-00:
I'm sorry, looks like I accidentally restricted your rights, please try now

avieth:
> I'm sorry, looks like I accidentally restricted your rights, please try now

All better now, thanks!

src/Node/Message/Store.hs
```diff
@@ -0,0 +1,61 @@
...
+    => Word32
+    -> BS.ByteString
+    -> Maybe BS.ByteString
+    -> Decoder t
+storeDecoderBody !remaining !acc !mbs = case mbs of
+    Nothing -> Fail BS.empty (fromIntegral (BS.length acc)) "Unexpected end of input (body)"
+    Just bs ->
+        let (front, back) = BS.splitAt (fromIntegral remaining) bs
+            taken = fromIntegral (BS.length front)
+            acc' = BS.append acc front
```
georgeee: Wouldn't it lead to quadratic time of parsing message (i.e. if we have many small-size chunks and big message)?

avieth: This will copy `acc` and `front` to a new buffer. Probably not ideal. Perhaps it'd be faster to create a builder and then build it right before calling `Store.decode`? @dcoutts

dcoutts: The simplest and fastest thing is to accumulate all the chunks and concatenate them once at the end.

PR #65, opened 2017-06-09
[CSL-738] Use most typical monad constraints in NTP

Martoon-00:
This is part of planned efforts to temporally get rid of
`Mockable Throw` & `Mockable Catch` usages.

Also `MonadMask` is removed, since `bracketOnError` can be performed
in terms of `MonadBaseControl IO`.


Martoon-00:
Can't argue against the last point :peka:

PR #66, opened 2017-06-12
[TW-150] added JSON logging

brunjlar:
Moved JSON logging from cardano-sl to time-warp-nt.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ cabal.project.local
 
 # vim related
 *~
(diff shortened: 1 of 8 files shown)


brunjlar:
@avieth @dcoutts @kantp The CI build kept failing, but I then edited the `travis` file to use one job instead of four, and it finally worked. Unfortunately, I have no write access, so could either of you please either grant me write access or merge the pull request?

PR #71, opened 2017-06-16
Export SendActions constructor

avieth:


avieth:
@georgeee @dcoutts tiny change just need an approval checkmark. cardano-sl uses the `SendActions` constructor. Maybe it shouldn't but we could fix that up later.

avieth:
> cardano-sl uses the SendActions constructor.

Sorry this was misleading. I'm patching CSL to use master time-warp-nt and I thought I'd have it use the standard `SendActions` type (it currently defines its own `SendActions` type with a slightly different `withConnectionTo` type). But changing this is out of scope of the work I'm doing so I'll close this up.

PR #73, opened 2017-06-17
[CSL-1209] Fix MonadBaseControl instance for Production

volhovM:
This fixes `MonadBaseControl` instance that was causing CSL-1209 bug. See more in YT description. TLDR it was delaying some execution when combined with `control`.


Martoon-00:
I merge, since it's crucial fix, and it doesn't make CI build worse than it currently is

PR #75, opened 2017-06-19
[CSL-1154] JsonLogConfig type

int-index:
In https://github.com/input-output-hk/cardano-sl/pull/967/ I removed as many transformer layers as possible, `JsonLogT` being one of them. Instead, I need to put this `JsonLogConfig` into a `ReaderT` at the bottom of the transformer stack, and then use `jsonLogDefault` to implement the `CanJsonLog` instance.

This patch does not change the semantics of the module and merely exports `JsonLogConfig` and `jsonLogDefault`.

(Ah, and I also fixed Haddock build)

### DIFF
--- a/src/Network/Broadcast/Relay/Class.hs
+++ b/src/Network/Broadcast/Relay/Class.hs
@@ -29,7 +29,7 @@ data Relay packingType m where
       , Msg.Message (ReqMsg key)
       , Msg.Message (InvOrData key contents)
       )
-      => (PropagationMsg packingType -> m ()) -- ^ How to relay the data.
+      => (PropagationMsg packingType -> m ()) -- How to relay the data.
       -> InvReqDataParams key contents m
       -> Relay packingType m
   Data ::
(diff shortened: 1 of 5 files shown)


stack.yaml
```diff
@@ -1,5 +1,9 @@
 resolver: lts-8.5
 
+flags:
+  ether:
+    disable-tup-instances: true
```
int-index: This speeds up CI.

src/Network/Broadcast/Relay/Class.hs
```diff
@@ -29,7 +29,7 @@ data Relay packingType m where
       , Msg.Message (ReqMsg key)
       , Msg.Message (InvOrData key contents)
       )
-      => (PropagationMsg packingType -> m ()) -- ^ How to relay the data.
```
gromakovsky: Did it break haddock? Why?

int-index: Yes, Haddock was broken because of these lines. I don't know the reason, it just didn't parse.

PR #76, opened 2017-06-20
[TW-148] relay broadcast

avieth:
It was a struggle to refactor cardano-sl to use a relay abstraction provided by time-warp. That's because CSL

1. Defines its own `SendActions` type.
2. Defines its own `Conversation` type.
3. Demands `OutSpecs` and `InSpecs` on workers and listeners.
4. Demands special constraints like `Bi` rather than `Serializable BiP`, and a `Message` constraint on the receive type of a `Conversation`.

After talking with Edsko and Philipp we decided that the relay listeners and logic etc. can all stay in CSL, and time-warp can simply provide a kind of outbound queueing system. That's what's given here in the `simpleRelayer`. We can make a proper type for a relayer (rather than a plain tuple of enqueue and dequeue) and then bring in Edsko's work to replace `simpleRelayer`.

This is the related patch to CSL https://github.com/input-output-hk/cardano-sl/pull/1043


dcoutts:
> TODO list:
>
>    Redesign the queueing mechanism so that it's possible to have different messages go into different queues, and so that one slow peer won't block all relays (currently it will!).

I'm sketching out things on this one.

PR #79, opened 2017-07-13
[TW-148] update enqueueConversation doc comment

avieth:
Writing credit: @dcoutts.

### DIFF
--- a/src/Node.hs
+++ b/src/Node.hs
@@ -164,12 +164,30 @@ data SendActions packing peerData peer msgClass m = SendActions {
               LL.NodeId
            -> (peerData -> Conversation packing m t)
            -> m t
-       -- Schedule a conversation with 0 or more peers.
-       -- After they have all been enqueued, a map is given with values which
-       -- will return once the conversation to that node has returned, possibly
-       -- throwing exceptions.
-       -- There is no required relationship between the input 'Set peer' and
-       -- the keys in the resulting 'Map peer (m t)'.
+
+       -- Schedule a conversation with some peers. The choice of /which/ peers
+       -- to converse with is primarily /not/ up to the caller, but is based on
+       -- the policy that the queue was set up with. The policy chooses the
+       -- peers to converse with based on the supplied message class (but also
+       -- on separate routing information and dynamic network conditions).
+       --
+       -- The parameter with a set of peers simply provides additional peers
+       -- that the policy can consider: they may or may not actually be used.
+       --
+       -- The action returns promptly, once the conversation has been placed in
+       -- the outbound queue. The resulting 'Map' contains an entry for each
+       -- peer that the conversation was actually enqueued for (i.e. the
+       -- outcome of the routing policy). It is possible for this map to be
+       -- empty, if no peer could be found to send to. This usually indicates a
+       -- failure.
+       --
+       -- The result 'Map' values are each an action to wait for the
+       -- conversation to be completed. They return the conversation result or
+       -- throw an exception if the conversation itself failed.
+       --
+       -- This way, if you ignore the 'Map' result then the enqueue is
+       -- asynchronous, but if you traverse the Map results then it is of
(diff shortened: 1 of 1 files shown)


edsko:
Ok, so I asked this on the other ticket before realizing that there was an open PR for this comment. Let me ask again, here:

> The result 'Map' values are each an action to wait for the conversation to be completed. They return the conversation result or throw an exception if the conversation itself failed.This way, if you ignore the 'Map' result then the enqueue is asynchronous, but if you traverse the Map results then it is of course synchronous.

Is that really true?

dcoutts:
Edsko is quite right. My docs are wrong, the current implementation is synchronous. This points out an issue: we need both sync and async enqueueConversation variants. For forwarding/relaying we need async (ie all the places in the old code that added things to the relay queue), where as all the places that waited for an answer need to be the sync interface.

The underlying outbound queue of course offers both. We'll need to extend the wrapper and the `SendActions`. While we're thinking about it, lets think of all the variants we need the `SendActions` wrappers for. These are all cases now provided by the outbound queue:

 * `enqueus`  ie async
 * `enqueueSync` or `enqueueSync'` which is what the current `SendActions` method uses
 * `enqueueCherished` which is what we'll use for edge nodes sending transactions
 * `enqueueTo` or `enqueueSyncTo` for sending requests for blocks (not sure if that needs to be sync or not, but we can do either)

dcoutts:
Oh, and as for the docs update, the first two paragraphs are ok but the last three can be replaced with:

> The action is synchronous, that is it returns once all of the conversations have been completed (either successfully of failure).
>
> The resulting 'Map' contains an entry for each peer that the conversation was actually done with. It is possible for this map to be empty, if no peer could be found to send to. This usually indicates a failure.
> 
> The result 'Map' values are each an action that return the conversation result or throw an exception if the conversation itself failed. So conversation exceptions are ignored unless you run the result actions (e.g. with 'traverse').

avieth:
> - enqueus ie async
> - enqueueSync or enqueueSync' which is what the current SendActions method uses

Both of these can be expressed by the same function, no? The existing `enqueueConversation` type captures it.

> - enqueueCherished which is what we'll use for edge nodes sending transactions
> - enqueueTo or enqueueSyncTo for sending requests for blocks (not sure if that needs to be sync or not, but we can do either)

Couldn't the special semantics of these be inferred within the `OutboundQ` by the sender and receiver `NodeType`s along with the message type?

edsko:
> Both of these can be expressed by the same function, no? The existing enqueueConversation type captures it.

Yes, _almost_. `enqueueAsync` now has the signature you need; the only difference between calling `enqueueAsync` and then calling every monadic action in the list to get the results, and `enqueueSync`, is that the latter does a bit more error checking and logs an error if _all_ sends failed. 

> Couldn't the special semantics of these be inferred within the OutboundQ by the sender and receiver NodeTypes along with the message type?

This is not so easy though. `enqueueCherished` is easily expressed in terms of `enqueueSync` but not so easily expressed in terms of `enqueueAsync`. I think this will need to remain to be a special case.

(I am also dubious that the signature that `enqueueConversation` wants, and the new `enqueueAsync` offers, is really the one we want; it may be useful for the integration effort right now but I think the `enqueue`/`enqueueSync` split that `OutboundQueue` offers is more intuitive.)

avieth:
> is that the latter does a bit more error checking and logs an error if all sends failed.

Maybe that's taking on too much responsibility. The caller can do that.

avieth:
> This is not so easy though. enqueueCherished is easily expressed in terms of enqueueSync but not so easily expressed in terms of enqueueAsync. I think this will need to remain to be a special case.

What's the difficulty? If `enqueueSync` can be expressed in terms of `enqueueAsync` then it's all good.

dcoutts:
This PR is completely stale now, the code has been changed around quite a bit since then.

PR #80, opened 2017-07-14
Add ability to enqueue to a specific set of nodes

edsko:
This set of nodes is still subject to the enqueueing policy. This PR also adds a new kind of message type called `MsgRequestData`. This new message type, along with the `enqueueTo` set of API calls, can be used when requesting data (such as blocks) from peers that are known to have them. The advantage is that the queue can then make an educated guess of _which_ of the peers-known-to-have-the-data we should request the data from; for example, we prefer to ask data from relay nodes over asking data from core nodes so that we don't put more load on the system when there are nodes in recovery mode.

This required some internal restructuring, primarily because we now have more than one kind of enqueueing instruction that can be returned by the enqueueing policy; the old `Enqueue` instruction is now called `EnqueueAll` ("enqueue to one alternative of all forwarding sets of the appropriate type"); the new one is `EnqueueOne` ("enqueue to one alternative of one forwarding set of any of the preferred types").

### DIFF
--- a/src/Network/Broadcast/OutboundQueue.hs
+++ b/src/Network/Broadcast/OutboundQueue.hs
@@ -49,11 +49,13 @@ module Network.Broadcast.OutboundQueue (
   , enqueue
   , enqueueSync'
   , enqueueSync
+  , enqueueAsync
   , enqueueCherished
     -- ** To specified peers
   , enqueueTo
   , enqueueSyncTo'
   , enqueueSyncTo
+  , enqueueAsyncTo
   , enqueueCherishedTo
     -- * Dequeuing
   , SendMsg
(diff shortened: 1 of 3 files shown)


avieth:
I don't see an enqueue variant which gives `m [(nid, m (Either SomeException t))]`. We'll need something from which `m (Map nid (m t))` can be derived, to fulfil the async `enqueueConversation` type.

edsko:
Ok, I can add one.

edsko:
@avieth Pushed a commit that adds `enqueueAsync` and `enqueueAsyncTo` with the signature you need.

avieth:
Sweeet. Will review now.

edsko:
@avieth @dcoutts So I see that `enqueue` now has the role of what was previously `enqueueAsync`, which makes sense to me. The only thing I'm wondering is, why in the demo did you replace `enqueue` with `enqueueSync'` in the forwarding thread? The old `enqueue` was asynchronous right (it simply ignored the results). Was this an intentional change?

PR #81, opened 2017-07-18
[TW-148] OutboundQ interface refinement

avieth:
The `OutboundQueue` abstraction is removed. Instead of building this into the `node` function and `SendActions`, the `SendActions` is also removed in favour of a `Converse` datatype (it's essentially the same thing as the `SendActions` from `master`).

Users who wish to use the `OutboundQ` (*not* to be confused with the `OutboundQueue`, which was removed) can use the `Converse` term made available in the `NodeAction` to spawn its `dequeueThread`. `Listener`s and `NodeAction` alike can use the `OutboundQ`'s `enqueue*` variants and `cherish` directly.

In the `OutboundQ` itself: the `Origin` term goes on the `MsgType` terms for which it's relevant, rather than as an extra argument to `enqueue`. The request blocks constructor has no `Origin`, but does take a `Peers` indicating to whom the request should be made.

Also removed the `enqueueTo*` variants. Instead, use the analogous `enqueue*` variant with `EnqueueToPeers peers` as the `EnqueueTo nid` argument. But there is no `EnqueueTo nid` argument on `enqueue*` variants. Instead, it's derived from the `MsgType nid`: requesting blocks enqueues to a particular peer set.


src/Network/Broadcast/OutboundQueue/Classification.hs
```diff
@@ -3,34 +3,64 @@
...
+    -- node; for a block it would mean it was constructed on this (core) node.
+    OriginSender
+
+    -- | It originated elsewhere; we're just forwarding it
+    --
+    -- We record on behalf of whom we're forwarding so that we can avoid
+    -- sending it straight back to them.
+  | OriginForward nid
+  deriving (Eq, Ord, Show)
+
```
dcoutts: Nice to have good docs!

src/Network/Broadcast/OutboundQueue/Classification.hs
```diff
@@ -3,34 +3,64 @@
...
 -- | Message types
-data MsgType =
+data MsgType nid =
```
dcoutts: Yep, nice. Nicely matches the structure of the type to the situation.

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -440,8 +443,19 @@ data OutboundQ msg nid = forall self .
 
       -- | Signal we use to wake up blocked threads
     , qSignal :: Signal CtrlMsg
+
+      -- | Mutable classification of peers.
+    , qClassifyPeers :: MVar (Classification nid)
```
dcoutts: Oh, it's a new MVar. Hmm, somehow I thought this would be integrated with the subscriptions. I need to think about this patch more carefully.

avieth: Is there ever a peer that we've classified, but which isn't subscribed to us? I think so but maybe I'm mistaken.

avieth: I suppose we could use `qSubscribers :: MVar (Peers nid, Classification nid)`.

edsko: As @avieth and I just discussed, the answer to "Is there ever a peer that we've classified, but which isn't subscribed to us? I think so but maybe I'm mistaken." is no: we will subscribe all statically known peers to the queue when we initialize it (this isn't done yet but will be soon).

edsko:
Should `EnqueueToParticular` be renamed? Maybe `EnqueueToSubsetOf`? We're not enqueueing to _all_ of those peers after all.

avieth:
> Should EnqueueToParticular be renamed? Maybe EnqueueToSubsetOf? We're not enqueueing to all of those peers after all.

Doesn't matter to me. Strong opinions about this @dcoutts @edsko ?

edsko:
I just think it's useful to be very explicit about what this does: it's not actually "send to these guys" but rather "pick one or more targets from". It's a minor thing, of course.

avieth:
I want to press the green button. Any objections? @dcoutts @edsko

avieth:
@edsko Just force pushed to resolve conflicts are #83 was merged. I squashed a bunch of my commits together but left your in tact.

edsko:
No objections from me.

avieth:
@edsko latest commit looks good. Ready to merge but will wait until @dcoutts gives it a final pass.

avieth:
@edsko @dcoutts I just pushed commit 195d21670aa8b3c08c354aef5796381e178ea7fd which eliminates empty `ByteString`s in `recvNext`. It's expected that the dispatcher will encounter empty input sometimes, for instance if the peer does `send conn []`.

PR #83, opened 2017-07-19
Redesign of packing type to support CBOR decoding

avieth:
- The `Decoder` type is parameterzsed by some `* -> *` type.
- A packing type need not (shouldn't) be inhabited, but must be an instance of the new `PackingType` class which determines concrete monadic context for encoding and decoding. For existing store and binary packing types these are all `Identity`.
- To use a packing type instance `packingType` to run a node over `m`, a `Packing packingType m` must be given. See `unpack` and `pack` in `Node.Message.Class`.
- Fix `recvNext` in case end of input happens during a partial parse. Previously it would `End` but now it correctly throws `NoParse`.
- `continueDecoding` now takes a single `ByteString` rather than `[ByteString]`.

### DIFF
--- a/src/Node.hs
+++ b/src/Node.hs
@@ -487,7 +487,7 @@ recvNext packing limit (LL.ChannelIn channel) = do
             unless (BS.null trailing) (Channel.unGetChannel channel (Just trailing))
             return outcome
   where
-    go remaining decoderStep = case decoderStep of
+    go !remaining decoderStep = case decoderStep of
         -- TODO use the error message in the exception.
         Fail _ _ _ -> throw NoParse
         Done trailing _ thing -> return (trailing, Input thing)
@@ -497,5 +497,5 @@ recvNext packing limit (LL.ChannelIn channel) = do
             case mbs of
                 Nothing -> runDecoder (next Nothing) >>= go remaining
                 Just bs ->
-                    let !remaining' = remaining - BS.length bs
+                    let remaining' = remaining - BS.length bs
                     in  runDecoder (next (Just bs)) >>= go remaining'
--- a/src/Node.hs
+++ b/src/Node.hs
@@ -84,8 +84,9 @@ import           Node.OutboundQueue
 import           Node.Internal              (ChannelIn, ChannelOut)
 import qualified Node.Internal              as LL
 import           Node.Message.Class         (Serializable (..), MessageName,
-                                             Message (..), messageName')
-import           Node.Message.Decoder       (Decoder (..), continueDecoding)
+                                             Message (..), messageName',
+                                             Packing, pack, unpack)
+import           Node.Message.Decoder       (Decoder (..), DecoderStep (..), continueDecoding)
 import           System.Random              (StdGen)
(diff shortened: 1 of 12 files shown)


avieth:
@dcoutts

src/Node.hs
```diff
@@ -493,7 +493,7 @@ recvNext packing limit (LL.ChannelIn channel) = do
             when (remaining <= 0) (throw LimitExceeded)
             mbs <- Channel.readChannel channel
             case mbs of
-                Nothing -> return (BS.empty, End)
+                Nothing -> go remaining (next Nothing)
                 Just bs ->
                     let !remaining' = remaining - BS.length bs
                     in  go remaining' (next (Just bs))
```
dcoutts: Not directly related to the fix, and just as a style recommendation, I would put the ! on the `go !remaining` argument, rather than using `let !` at the call site, ie

    go !remaining decoder = ...
        ....
             case mbs of
                 Nothing -> go remaining  (next Nothing)
                 Just bs -> go remaining' (next (Just bs))
                     where remaining' = remaining - BS.length bs

The advantage is that we only have to put the ! at the one definition site rather than the potentially multiple call sites. It also means when reading we can see immediately that the `remaining` is strict rather than having to scan the whole body. It also means there's nothing to think about if refactoring to inline or un-inline a let/where clause, e.g.

                 Just bs -> go (remaining - BS.length bs) (next (Just bs))

src/Node.hs
```diff
@@ -493,7 +493,7 @@ recvNext packing limit (LL.ChannelIn channel) = do
             when (remaining <= 0) (throw LimitExceeded)
             mbs <- Channel.readChannel channel
             case mbs of
-                Nothing -> return (BS.empty, End)
+                Nothing -> go remaining (next Nothing)
```
dcoutts: yep, great.

adinapoli-iohk:
On it as we speak guys, I hope to be able to give you the final 👍  as soon as I can!

src/Node.hs
avieth: Agreed

adinapoli-iohk:
@avieth @dcoutts This is the initial, horrid prototype:

https://github.com/adinapoli-iohk/cardano-sl/commit/cb8ee2389794f8188a9e9562614364f79152227d

Let me know if it makes any kind of sense to you!

avieth:
Weird CI failure... looks like everything went OK but then it timed out, even after the test suite was successfully run.

avieth:
> Let me know if it makes any kind of sense to you!

Makes perfect sense! Is everything working well with the revision I sent you? Even after this is merged, we'll still have to keep that special revision until #1116 of cardano-sl lands.

adinapoli-iohk:
> Makes perfect sense! Is everything working well with the revision I sent you?

It does! I had to make only one minor tweak to my instance which was generalise `bipPacking` to run in `MonadIO m` so that we can now use it in places like `Pos.Launcher.Runner.runServer`. This is ofc completely unrelated to your work but I thought I would share it. In a nutshell, great success indeed! 😀

PR #84, opened 2017-07-28
[CSL-849] messageCode :: Word16 to identify conv

avieth:
This is the simplest way to satisfy CSL-849, but I'm tempted to do more.

The message code / name should not be determined by the send type of the
conversation. Really what we want to express is a code for a listener.

### DIFF
--- a/src/Node/Message/Class.hs
+++ b/src/Node/Message/Class.hs
@@ -19,49 +19,26 @@ module Node.Message.Class
     , unpack
 
     , Message (..)
-    , messageName'
+    , messageCode'
 
-    , MessageName (..)
+    , MessageCode
     ) where
 
-import qualified Data.Binary                   as Bin
-import qualified Data.ByteString               as BS
 import qualified Data.ByteString.Lazy          as LBS
-import           Data.Data                     (Data, dataTypeName, dataTypeOf)
-import           Data.Hashable                 (Hashable)
-import           Data.Proxy                    (Proxy (..), asProxyTypeOf)
-import           Data.String                   (IsString, fromString)
+import           Data.Proxy                    (Proxy (..))
 import qualified Data.Text                     as T
-import           Data.Text.Buildable           (Buildable)
-import qualified Data.Text.Buildable           as B
+import           Data.Word                     (Word16)
 import qualified Formatting                    as F
-import           GHC.Generics                  (Generic)
-import           Serokell.Util.Base16          (base16F)
 import           Node.Message.Decoder          (Decoder, hoistDecoder)
 
 -- * Message name
 
-newtype MessageName = MessageName BS.ByteString
-deriving instance Eq MessageName
-deriving instance Ord MessageName
-deriving instance Show MessageName
-deriving instance Generic MessageName
-deriving instance IsString MessageName
-deriving instance Hashable MessageName
-deriving instance Monoid MessageName
-instance Bin.Binary MessageName
+type MessageCode = Word16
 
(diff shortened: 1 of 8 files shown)


src/Node/Message/Class.hs
```diff
@@ -34,13 +35,16 @@ import           Data.String                   (IsString, fromString)
...
 import qualified Formatting                    as F
 import           GHC.Generics                  (Generic)
 import           Serokell.Util.Base16          (base16F)
 import           Node.Message.Decoder          (Decoder, hoistDecoder)
 
 -- * Message name
 
+type MessageCode = Word16
+
 newtype MessageName = MessageName BS.ByteString
```
georgeee: What's purpose of not removing `MessageName`?

georgeee: What's purpose of not removing `MessageName`?

avieth: Yes I suppose we ought to remove it. Will do that

avieth:
42d2b9c3347b6cca0098a0b51e47bbf07456b73f minimally satisfies the original ticket. `MessageName` is removed in favour of `MessageCode ~ Word16` and the `Message` class associates a `MessageCode` with a type (in practice, the "send" type of a conversation).

## Integration with cardano-sl

I've been studying cardano-sl to see how we might do this, and found no technical hurdles, but it won't be a very pretty change either. Currently there's a `MessagePart` class, which is used to derive `Message` instances for the inv/req/data message types (`DataMsg`, `InvOrData`, etc.) by `<>`ing the `MessageName` of the latter with the one provided by the `MessagePart` instance of the former. I see why this was done: it prevents instance multiplication. Switching to `MessageCode`, we *could* do something similar, perhaps by using the top 8 bits to identify the inv/req/data message, and the low-8 bits to identify the type of the thing it carries. But anyway, I don't think we should go down that road; I think it would be better to state each `MessageCode` explicitly because they define the protocol and we'll want a reference for them. So we'll see a proliferation of `Message` instances in `src/Pos/Communication/Message.hs` but I think it's better that way. Agree/disagree? @georgeee 

Although bringing this in would technically satisfy CSL-849, it wouldn't eliminate the underlying problem, because we still pull in `HashMap`s as part of the `VerInfo` and these can be arbitrarily large. Those are for the `OutSpecs`/`InSpecs`. Shall we leave that for another ticket?

Also: the number of `Message` instances required is artificially inflated by the fact that both types used in a conversation require a `MessageCode`, for the benefit of `OutSpecs`/`InSpecs`. Could we possibly relax that? Wouldn't it be enough to state just the `MessageCode` of each conversation that we accept (have a listener for) and may use (a worker will try to establish a conversation there)? In other words, does the `ConvHandler` constructor of `HandlerSpec` really need to have a `MessageCode`?

## A moral correction

This is perhaps beyond the scope of CSL-849 but I do want to bring it up.

The `Message` class really shouldn't exist (at least not in time-warp-nt) and the way it's used is in my opinion confusing and just wrong (though admittedly it gets the job done). **The type of the thing that's sent identifies the conversation**. I have a patch that gets rid of this constraint. A `ConversationId ~ Word16` is given when a conversation is to be established, and is also given as part of the listener index when a node is brought up, and a conversation is able to send/recv any serializable type:

```
send :: Serializable packing snd => ConversationActions packing m -> snd -> m ()
recv :: Serializable packing rcv => ConversationActions packing m -> Word32 -> m (Maybe rcv)
```

This is definitely beyond the scope of CSL-849 but I think it should be brought in eventually.

PR #85, opened 2017-08-02
Generalize to multiple buckets

edsko:
The set of buckets is static and ordered, which simplifies reasoning about the 'Peers' values we're working with, as well as integration in the settlement layer.

### DIFF
--- a/src/Network/Broadcast/OutboundQueue.hs
+++ b/src/Network/Broadcast/OutboundQueue.hs
@@ -60,9 +60,7 @@ module Network.Broadcast.OutboundQueue (
   , Alts
   , simplePeers
   , peersFromList
-  , updateKnownPeers
-  , addKnownPeers
-  , removeKnownPeer
+  , updatePeersBucket
     -- * Debugging
   , dumpState
   ) where
@@ -75,7 +73,6 @@ import Control.Monad.IO.Class
 import Data.Either (rights)
 import Data.Maybe (maybeToList)
 import Data.Map.Strict (Map)
-import Data.Monoid ((<>))
 import Data.Set (Set)
 import Data.Text (Text)
 import Data.Time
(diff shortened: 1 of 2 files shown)


src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -1032,32 +1044,33 @@ flush OutQ{..} = liftIO $ do
...
+-- NOTE: Any messages to peers that no longer exist in _any_ bucket will be
+-- removed from the queue.
+updatePeersBucket :: MonadIO m
+                  => OutboundQ msg nid buck
+                  -> buck
+                  -> (Peers nid -> Peers nid)
+                  -> m ()
+updatePeersBucket outQ@OutQ{..} buck f = do
+    before <- getAllPeers outQ
+    let Bucket bucket = qBuckets Map.! buck
```
edsko: I started with qBuckets being an `MVar` but I removed it: it's much easier to understand the behaviour of the queue when the set of buckets is entirely static (and ordered).

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -1032,32 +1044,33 @@ flush OutQ{..} = liftIO $ do
...
+    applyMVar_ bucket f
+    after <- getAllPeers outQ
+    let removed = peersToSet before Set.\\ peersToSet after
```
edsko: I'm not sure this is something we can prevent. If there is a race condition between a thread removing a node and another one introducing it, it's not entirely clear what the final outcome will be anyway. What is the invariant you would like to have?

avieth: > it's not entirely clear what the final outcome will be anyway

Depending on which thread enters `updatePeersBucket` first: 

- The node is added (re-added)  to the buckets, and then subsequently removed by the other thread.
- The node is *completely* removed (relevant updates to `qInFlight`, `qFailures`, and the `MQ` finish), and the node is subsequently re-added

edsko:
I've given this some more thought and I think I see what you mean now; I don't think it's necessarily _worse_ than the situation before we introduced multiple buckets, but it's true that we can now do better.

Consider the situation as it was before we had multiple buckets. We have two threads racing: thread _A_ removes node _n_, and thread _B_ adds node _n_ and enqueues a message to _n_. The interleaving that you're worried about it

```
A                    B
remove n,       
compute `removed`
                     reintroduce n,
                     enqueue message to n
dequeue all 
messages to n                     
```

where we end up in a situation with `n` registered, but a message to `n` lost. Admittedly, this isn't a great scenario (since it violates linearizability of `updatePeersBucket`), but my point above was: if we have two threads racing, one removing `n` and one reintroducing `n`, we already have the following  interleaving also

```
                    introduce n (no-op: it's already there),
                    enqueue message to n
remove n, 
dequeue all 
messages to n
```

This is a perfectly linearizable interleaving (in that sense one that would still be possible if `updatePeersBucket` was atomic) I'm not sure that the outcome is so much better from the point of view of the next layer up.

However, now that we have multiple buckets, the above linearizable interleaving is no longer possible: when _B_ introduces _n_ into its bucket, _A_ can then remove _n_ from it's own bucket all it wants, _n_ will still be considered to be part of the total set of all peers and hence we will not dequeue any messages.

So the property we would like to have, I think, and can now guarantee, is that

**Invariant: if thread _T_ adds node _n_ to its peers bucket, and then enqueues a message to it, that message should not be deleted.**

We can guarantee this provided that each bucket is updated by a single thread only; and indeed, the easiest way to do that is to have a single MVar `qBuckets`, just like you suggested (I deserve an "I told you so" :-) ).

PR #87, opened 2017-08-07
EKG counters

edsko:
https://issues.serokell.io/issue/CSL-1462
https://issues.serokell.io/issue/CSLREQ-198
https://issues.serokell.io/issue/CSL-1465

### DIFF
--- a/QUEUE_METRICS.md
+++ b/QUEUE_METRICS.md
@@ -109,9 +109,9 @@ we would like to have; the queue will then enqueue the message to _one_ node
 in this set; we call this an "enqueue-one instruction".
 
 As for enqueue-all instructions, an enqueue-one instruction is considered to
-have failed when we could not _any_ node to enqueue the message to (that is,
-any node in the specified set with no recent failures). In this case the message
-is lost.
+have failed when we could not find _any_ node to enqueue the message to (that
+is, any node in the specified set with no recent failures). In this case the
+message is lost.
 
 ### Failure to choose an alternative (e.g., `queue.node0.FailedChooseAlt`)
 
--- a/QUEUE_METRICS.md
+++ b/QUEUE_METRICS.md
@@ -1,0 +1,159 @@
+# EKG metrics for the OutboundQueue
+
+The outbound queue introduces a number of EKG metrics that can be used used to
+monitor the queue's health. There are a number of guages showing current status
+and a number of counters that keep track of error conditions.
+
+## Guages
+
+### InFlight (e.g. `queue.node0.InFlight`)
+
+Records the total number of conversations that are currently on-going with
+peers.
+
+### Failures (e.g., `queue.node0.Failures`)
+
+Records the total number nodes that had a recent failure (the conversion
+threw an exception). We will not attempt to enqueue messages to such nodes.
+
+### Scheduled (e.g., `queue.node0.Scheduled`)
+
+Records the total number of conversations that have been scheduled (and
+enqueued to specific peers), but not yet started.
+
+### Known peers (e.g., `queue.node0.BucketStatic`)
+
+The outbound queue keeps a number of buckets of peers it knows about.
+
+* `BucketStatic`: statically known peers
+  (from a `topology.yaml` file; core and relay nodes only).
+  This should only change on `SIGHUP` and remain static otherwise.
+
+* `BucketSubscriptionListener`: subscribers
+  (peers that subscribed to this node by sending a `MsgSubscribe` message).
+  A high number indicates a (relay) node under heavy load.
+
+* `BucketBehindNatWorker`: relay nodes discovered through DNS query
+  (used by behind-NAT edge nodes to declare relay nodes as known peers).
+  This should be equal to the number of relay nodes returned by DNS
+  (through a series of DNS queries to the various hostnames).
+
+* `BucketKademliaWorker`: peers discovered through the Kademlia network
+  (used in P2P mode).
+  This number should remain more or less constant (configurable through the
+  `topology.yaml` file), unless there are not enough Kademlia peers to discover.
+
+For each of these buckets we counts the number of distinct known peers (i.e.,
+ignoring routing information).
+
+**Note on directionality**: The outbound queue deals with outgoing messages
+only. Thus, the situation looks like this for behind-NAT nodes:
+
+```
+  /-------------------------\           /------------------------------\
+  |                         |           |                              |
+  |  BEHIND-NAT NODE        |<----------+--BucketSubscriptionListener  |
+  |                         |           |                              |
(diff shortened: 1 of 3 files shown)


QUEUE_METRICS.md
```diff
@@ -0,0 +1,159 @@
...
+### Failed "enqueue-one" instructions (e.g., `queue.node0.FailedEnqueueOne`)
+
+Occassionally the queue is provided with a message that should be sent to one
+of a specific set of nodes. Typically this happens when we learn that a certain
+set of nodes has some information (for instance, a particular block) that
+we would like to have; the queue will then enqueue the message to _one_ node
+in this set; we call this an "enqueue-one instruction".
+
+As for enqueue-all instructions, an enqueue-one instruction is considered to
+have failed when we could not _any_ node to enqueue the message to (that is,
```
dcoutts: could not **find** any

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -503,6 +515,180 @@ new qSelf qEnqueuePolicy qDequeuePolicy qFailurePolicy = liftIO $ do
     return OutQ{..}
 
 {-------------------------------------------------------------------------------
+  Monitoring the queue's health
+-------------------------------------------------------------------------------}
+
+-- | Queue health metrics
+data QHealth = QHealth {
```
dcoutts: Why not EKG counters directly? We already depend on EKG directly in this module, and these are write-only counters right (unlike others that are stats used for enqueue/dequeue logic)?

Or if not EKG counters, why not IORefs?

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -503,6 +515,180 @@ new qSelf qEnqueuePolicy qDequeuePolicy qFailurePolicy = liftIO $ do
...
+--
+-- The third type argument indicates the additional information we need to
+-- format a human-readable error or warning message for the failure.
+data Failure :: (* -> *) -> * -> * -> * where
+    FailedEnqueueAll  :: Failure msg nid (Enqueue, Some msg, [Alts nid])
+    FailedEnqueueOne  :: Failure msg nid (Enqueue, Some msg, [(NodeType, Alts nid)])
+    FailedAllSends    :: Failure msg nid (Some msg, [nid])
+    FailedCherishLoop :: Failure msg nid ()
+    FailedChooseAlt   :: Failure msg nid [nid]
+    FailedSend        :: Failure msg nid (Some (Packet msg nid), SomeException)
```
dcoutts: Nice way of factoring things.

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -503,6 +515,180 @@ new qSelf qEnqueuePolicy qDequeuePolicy qFailurePolicy = liftIO $ do
...
+    --
+    -- Since these buckets are added to the map when needed, but we register
+    -- all gauges up-front, we need to enumerate the list of buckets
+    --
+    -- NOTE: By reporting this per bucket, we can easily distinguish the number
+    -- of subscribers ('BucketSubscriptionListener') from, say, statically known
+    -- peers ('BucketStatic').
+    forM_ [minBound .. maxBound] $ \buck ->
+      regGauge [show buck] $
+        countPeers . Map.findWithDefault mempty buck <$> readMVar qBuckets
```
dcoutts: Nice.

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -1005,19 +1218,32 @@ enqueueCherished outQ msgType msg =
...
+    peers <- restrict =<< getAllPeers outQ
+    intEnqueue outQ msgType msg peers
   where
-    restriction = case enqTo of
-      EnqueueToAll -> id
-      EnqueueToSubset peers' -> restrictPeers peers'
+    -- Restrict the set of all peers to the requested subset (if one).
+    --
+    -- Any unknown peers will be used as a fallback in case no other peers
+    -- can be found.
```
avieth: Is this controllable by policy? `MsgGetBlocks` uses the `EnqueueToSubset` restriction and I wonder if it's ever a good idea for, say, a relay to make the request to unknown peers.

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -1186,27 +1406,46 @@ getAllPeers OutQ{..} = liftIO $ fold <$> readMVar qBuckets
...
+
+    unless success $
+      logFailure outQ (FailedBucketFull buck) ()
+    return success
   where
     f' :: Maybe (Peers nid) -> Maybe (Peers nid)
     f' Nothing      = Just $ f mempty
     f' (Just peers) = Just $ f peers
 
+data PeersBucketSizeExceeded = PeersBucketSizeExceeded
```
avieth: Unused datatype?

src/Network/Broadcast/OutboundQueue.hs
edsko: The relays _will_ send to unknown peers because the static topology will not be bidirectional.

src/Network/Broadcast/OutboundQueue.hs
edsko: Quite.

src/Network/Broadcast/OutboundQueue.hs
avieth: Oh I see, they'll know the classification from the shared topology file, but the peer won't be in the `Peers` value.

src/Network/Broadcast/OutboundQueue.hs
avieth: I pushed a commit to remove it.

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -1178,19 +1218,32 @@ enqueueCherished outQ msgType msg =
...
+    intEnqueue outQ msgType msg peers
   where
-    restriction = case enqTo of
-      EnqueueToAll -> id
-      EnqueueToSubset peers' -> restrictPeers peers'
+    -- Restrict the set of all peers to the requested subset (if one).
+    --
+    -- Any unknown peers will be used as a fallback in case no other peers
+    -- can be found.
+    restrict :: Peers nid -> m (Peers nid)
```
dcoutts: This function is actually pure, doesn't need to be in `m`.

avieth:
> LGTM so far. Have only reviewed the first patch so far since the second was WIP at the time of review and I didn't grok it yet.

LGTM. @dcoutts what say you?

src/Network/Broadcast/OutboundQueue.hs
avieth: done

PR #88, opened 2017-08-09
Add clearRecentFailures

edsko:


edsko:
@avieth @dcoutts Made this PR against master, but cardano-sl is not currently using `master`, but `time-warp-nt-0.2`. I could rebase against that but then we'd need to make sure we don't lose it when we do move to `master`.

edsko:
Actually, scratch that. I can't test it now since cardano-sl does'nt build against master. I'll rebase this against time-warp-nt-0.2 and then we'll need to resolve this.

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -815,6 +816,13 @@ intFailure OutQ{..} threadRegistry@TR{} p sendExecTime err = do
 hasRecentFailure :: MonadIO m => OutboundQ msg nid buck -> nid -> m Bool
 hasRecentFailure OutQ{..} nid = liftIO $ Set.member nid <$> readMVar qFailures
 
+-- | Reset internal statistics about failed nodes
+--
+-- This is useful when we know for external reasons that nodes may be reachable
+-- again, allowing the outbound queue to enqueue messages to those nodes.
+clearRecentFailures :: MonadIO m => OutboundQ msg nid buck -> m ()
+clearRecentFailures OutQ{..} = applyMVar_ qFailures $ const Set.empty
```
avieth: Should kill any threads which are waiting to delete an element from this set. If we don't, they could prematurely clear subsequent failures.

avieth:
cardano-sl won't build against time-warp-nt master because this https://github.com/input-output-hk/cardano-sl/pull/1288 hasn't been merged yet. Sorry for the delay, CI is having issues.

src/Network/Broadcast/OutboundQueue.hs
edsko: I figured it didn't matter, but that's a good point. Will fix.

edsko:
Ok, we now record recent failures in a different manner, not relying on threads at all. This addresses @avieth 's concern, I think. Also rebased against `master` since https://github.com/input-output-hk/cardano-sl/pull/1288 has been merged.

PR #89, opened 2017-08-11
Bring policies in line with latest spec / bugfixes

edsko:
This is a crucial PR, because in the policy as it had been implemented requests for block headers and requests for blocks were given the lowest priority. This would cause relay nodes that are inundated with transactions (which were given a higher priority) to never request blocks, therefore getting behind, finally entering recovery mode, and then being unable to exit recovery mode because transactions were also given higher priority than the "request headers" messages.
    
The original spec did not talk about these kinds of messages, but the newer specs do and indeed give them high priority. This PR brings the code in line with this spec, but there are still some open questions:
    
* I have increased the `MaxInFlight` values for core-to-relay, relay-to-core, and edge-to-relay communication to a minumum of 2. Otherwise we may run into deadlocks, as described in https://github.com/input-output-hk/cardano-sl/pull/1323 . That PR addresses the problem higher up the chain, but I worry that the same problem will arise elsewhere (in fact, we may wish to have a slightly higher minimum than 2 even).

* The policy docs do not give rules for header requests and block requests for edge nodes, and moreover gives transactions the highest priority (made sense, since only message considered).  This opens it up to the same problems as described above, however; therefore, I have given transactions a `PLow` transaction (P4), with requests for block headers and requests for blocks a `PHigh` priority (P2). This brings the policy here in line with core and relay nodes. For exchanges the policy gave P1 for header announcements and P2 for transactions; I have left the P1 as is, but here too made the transaction P4 and the requests for headers/blocks P2.
    
* I am worried about the low max-ahead value of 1 for requests for blocks and requests for headers. What happens when a node enters recovery mode, and tries to enqueue a whole series of block requests because it realizes it is far behind? It will quickly run out of peers to encode the requests to. Maybe this is what the "fallback: TODO" is about in the policy document. Therefore, I have set the max-ahead for block requests to infinite. I have however also makes a related change: when we have a choice, we pick a node with the smallest number of messages ahead. This means we at least get some load balancing when we do request a ton of blocks.

This PR also makes some minor fixes to logging; as @kantp and I were debugging, we discovered some minor inconsistencies in the logging that were confusing us.

### DIFF
--- a/src/Network/Broadcast/OutboundQueue.hs
+++ b/src/Network/Broadcast/OutboundQueue.hs
@@ -140,7 +140,7 @@ enumPrecHighestFirst = reverse enumPrecLowestFirst
 --
 -- If we cannot find any alternative that doesn't match requirements we simply
 -- give up on forwarding set.
-data MaxAhead = MaxAhead Int
+newtype MaxAhead = MaxAhead Int
   deriving Show
 
 -- | Enqueueing instruction
--- a/src/Network/Broadcast/OutboundQueue.hs
+++ b/src/Network/Broadcast/OutboundQueue.hs
@@ -31,12 +31,22 @@ module Network.Broadcast.OutboundQueue (
   , Enqueue(..)
   , EnqueuePolicy
   , defaultEnqueuePolicy
+  , defaultEnqueuePolicyCore
+  , defaultEnqueuePolicyRelay
+  , defaultEnqueuePolicyEdgeBehindNat
+  , defaultEnqueuePolicyEdgeExchange
+  , defaultEnqueuePolicyEdgeP2P
     -- ** Dequeueing policy
   , RateLimit(..)
   , MaxInFlight(..)
   , Dequeue(..)
   , DequeuePolicy
   , defaultDequeuePolicy
+  , defaultDequeuePolicyCore
+  , defaultDequeuePolicyRelay
+  , defaultDequeuePolicyEdgeBehindNat
+  , defaultDequeuePolicyEdgeExchange
+  , defaultDequeuePolicyEdgeP2P
     -- ** Failure policy
   , FailurePolicy
   , ReconsiderAfter(..)
@@ -159,40 +169,47 @@ data Enqueue =
 -- of the next layer up to configure these peers as desired.
 --
 -- TODO: Sanity check the number of forwarding sets and number of alternatives.
--- TODO: Verify the max queue sizes against the updated policy.
 type EnqueuePolicy nid =
            MsgType nid  -- ^ Type of the message we want to send
         -> [Enqueue]
 
--- TODO: Verify the policy for data requests
-defaultEnqueuePolicy :: NodeType           -- ^ Type of this node
-                     -> EnqueuePolicy nid
-defaultEnqueuePolicy NodeCore = go
+-- | Pick default policy given node type
+--
+-- NOTE: Assumes standard behind-NAT node in the case of edge nodes.
+defaultEnqueuePolicy :: NodeType -> EnqueuePolicy nid
+defaultEnqueuePolicy NodeCore  = defaultEnqueuePolicyCore
+defaultEnqueuePolicy NodeRelay = defaultEnqueuePolicyRelay
+defaultEnqueuePolicy NodeEdge  = defaultEnqueuePolicyEdgeBehindNat
+
+-- | Default enqueue policy for code nodes
+defaultEnqueuePolicyCore :: EnqueuePolicy nid
+defaultEnqueuePolicyCore = go
   where
-    -- Enqueue policy for core nodes
     go :: EnqueuePolicy nid
     go (MsgAnnounceBlockHeader _) = [
         EnqueueAll NodeCore  (MaxAhead 0) PHighest
-      , EnqueueAll NodeRelay (MaxAhead 0) PMedium
+      , EnqueueAll NodeRelay (MaxAhead 0) PHigh
       ]
     go MsgRequestBlockHeaders = [
-        EnqueueAll NodeCore  (MaxAhead 20) PLowest
-      , EnqueueAll NodeRelay (MaxAhead 20) PLowest
+        EnqueueAll NodeCore  (MaxAhead 1) PHigh
+      , EnqueueAll NodeRelay (MaxAhead 1) PHigh
       ]
-    -- TODO use the argument (Set nid).
     go (MsgRequestBlock _) = [
         -- We never ask for data from edge nodes
-        EnqueueOne [NodeRelay, NodeCore] (MaxAhead 20) PLowest
+        EnqueueOne [NodeRelay, NodeCore] (MaxAhead 1) PHigh
       ]
     go (MsgMPC _) = [
-        EnqueueAll NodeCore (MaxAhead 1) PHigh
+        EnqueueAll NodeCore (MaxAhead 1) PMedium
         -- not sent to relay nodes
       ]
     go (MsgTransaction _) = [
         EnqueueAll NodeCore (MaxAhead 20) PLow
         -- not sent to relay nodes
       ]
-defaultEnqueuePolicy NodeRelay = go
+
+-- | Default enqueue policy for relay nodes
+defaultEnqueuePolicyRelay :: EnqueuePolicy nid
+defaultEnqueuePolicyRelay = go
   where
     -- Enqueue policy for relay nodes
     go :: EnqueuePolicy nid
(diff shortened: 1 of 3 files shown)


avieth:
> I have increased the MaxInFlight values for core-to-relay, relay-to-core, and edge-to-relay communication to a minumum of 2. Otherwise we may run into deadlocks, as described in input-output-hk/cardano-sl#1323 . That PR addresses the problem higher up the chain, but I worry that the same problem will arise elsewhere (in fact, we may wish to have a slightly higher minimum than 2 even).

Can't we say for certain that it will never arise, so long as we don't block on enqueued conversations *within* other conversations?

edsko:
> Can't we say for certain that it will never arise, so long as we don't block on enqueued conversations within other conversations?

Let me know when you've audited the Cardano layer for that :)

avieth:
> Let me know when you've audited the Cardano layer for that :)

Surprisingly, this will be rather easy to do.

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -232,12 +253,63 @@ defaultEnqueuePolicy NodeEdge = go
         -- not forwarded
       ]
     go MsgRequestBlockHeaders = [
-        EnqueueAll NodeRelay (MaxAhead 20) PLowest
+        EnqueueAll NodeRelay (MaxAhead 1) PHigh
```
dcoutts: This should be max ahead 0. Since we don't send header announcements then those are not expected to be ahead of us (as they are in the case of other node types).

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -232,12 +253,63 @@ defaultEnqueuePolicy NodeEdge = go
...
+defaultEnqueuePolicyEdgeP2P = go
+  where
+    -- Enqueue policy for edge nodes
+    go :: EnqueuePolicy nid
+    go (MsgTransaction OriginSender) = [
+        EnqueueAll NodeRelay (MaxAhead 3) PLow
+      ]
+    go (MsgTransaction (OriginForward _)) = [
+        -- don't forward transactions that weren't created at this node
+      ]
```
dcoutts: Actually for P2P nodes we'll do exactly that, so just use  `MsgTransaction _` rather than `MsgTransaction OriginSender` in the first case above and omit the second case.

avieth:
> I'm a little nervous of the UnlimitedMaxAhead. I'd like to check with Alex or George about whether block retrieval uses one of many conversations for actually fetching blocks. If it does, I wonder if we should bump down the priority by one since if we have both block header requests and block requests at the same priority level and there can be many block requests, but the block header requests use max ahead 1 then we would have a problem.

In `src/Pos/Block/Network/Retrieval.hs`, `handleCHsValid` will use `retrieveBlocks` within a conversation, which will get all the needed blocks. So it's one unit as far as the outbound queue is concerned.

avieth:
We shouldn't need an `UnlimitedMaxAhead`, because one get blocks conversation will retrieve blocks for multiple headers.

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -159,40 +171,47 @@ data Enqueue =
...
+      , EnqueueAll NodeRelay (MaxAhead 0) PHigh
       ]
     go MsgRequestBlockHeaders = [
-        EnqueueAll NodeCore  (MaxAhead 20) PLowest
-      , EnqueueAll NodeRelay (MaxAhead 20) PLowest
+        EnqueueAll NodeCore  (MaxAhead 1) PHigh
+      , EnqueueAll NodeRelay (MaxAhead 1) PHigh
       ]
-    -- TODO use the argument (Set nid).
     go (MsgRequestBlock _) = [
```
avieth: Shouldn't cut out that TODO, it's important! We need a new variant of `Enqueue`: `EnqueueParticular` which chooses a given set of peers and tries to hit all of them. Without this, a node will request blocks *not necessarily* from the same node that it took the headers from.

dcoutts:
> We shouldn't need an UnlimitedMaxAhead, because one get blocks conversation will retrieve blocks for multiple headers.

Ok, so you mean that Edsko's concern here is unfounded because it's based on the assumption that we enqueue lots of separate get blocks conversations in one go, whereas in fact we enqueue exactly one at once right?

> I am worried about the low max-ahead value of 1 for requests for blocks and requests for headers. What happens when a node enters recovery mode, and tries to enqueue a whole series of block requests because it realizes it is far behind?

avieth:
> Ok, so you mean that Edsko's concern here is unfounded because it's based on the assumption that we enqueue lots of separate get blocks conversations in one go, whereas in fact we enqueue exactly one at once right?

Exactly.

src/Network/Broadcast/OutboundQueue.hs
dcoutts: Actually this TODO and the other equivalent TODOs are in fact done. The `EnqueueOne [NodeRelay, NodeCore]` bit is exactly what says that we have to send to one of the supplied node ids (though preferring relays over core nodes and not permitting requests to edge nodes except perhaps by fallback behaviour).

avieth:
> > Ok, so you mean that Edsko's concern here is unfounded because it's based on the assumption that we enqueue lots of separate get blocks conversations in one go, whereas in fact we enqueue exactly one at once right?
> 
> Exactly.

Well, right now there's a bug: `MsgGetBlocks thesePeers` will ignore `thesePeers :: Set NodeId` and instead enqueue to whatever set of peers is best. So it could enqueue a conversation to more than one node, but yes for each peer it enqueues *at most* one conversation which will get all the blocks it needs.

src/Network/Broadcast/OutboundQueue.hs
avieth: The argument is a set of particular node identifiers. `EnqueueOne [NodeRelay, NodeCore]` doesn't pick any particular node, it picks by node type.

dcoutts:
> Well, right now there's a bug: MsgGetBlocks thesePeers will ignore thesePeers :: Set NodeId and instead enqueue to whatever set of peers is best. So it could enqueue a conversation to more than one node, but yes for each peer it enqueues at most one conversation which will get all the blocks it needs.

I think that's not correct, all occurrences in the policy of `MsgRequestBlock` use the `EnqueueOne` policy type. This is exactly the policy type that uses the `thesePeers :: Set NodeId` argument.

avieth:
I see the `MsgRequestBlock` argument indeed is used, via `msgEnqueueTo`.

src/Network/Broadcast/OutboundQueue.hs
avieth: Resolved. `msgEnqueueTo` restricts the set of peers so it's all good!

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -253,11 +253,11 @@ defaultEnqueuePolicyEdgeBehindNat = go
...
       ]
     go MsgRequestBlockHeaders = [
-        EnqueueAll NodeRelay (MaxAhead 1) PHigh
+        EnqueueAll NodeRelay (MaxAhead 0) PHigh
       ]
-    go (MsgRequestBlock _) = [
+    go (MsgRequestBlocks _) = [
         -- Edge nodes can only talk to relay nodes
-        EnqueueOne [NodeRelay] UnlimitedMaxAhead PHigh
+        EnqueueOne [NodeRelay] (MaxAhead 0) PHigh
```
edsko: Why 0 here but 1 above?

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -295,21 +295,18 @@ defaultEnqueuePolicyEdgeP2P = go
...
         EnqueueAll NodeRelay (MaxAhead 0) PHighest
       ]
     go MsgRequestBlockHeaders = [
         EnqueueAll NodeRelay (MaxAhead 1) PHigh
       ]
-    go (MsgRequestBlock _) = [
+    go (MsgRequestBlocks _) = [
         -- Edge nodes can only talk to relay nodes
-        EnqueueOne [NodeRelay] UnlimitedMaxAhead PHigh
+        EnqueueOne [NodeRelay] (MaxAhead 1) PHigh
```
edsko: .. and 1 here again ..

src/Network/Broadcast/OutboundQueue.hs
avieth: Took it from a doc posted in the timewarp channel.

avieth:
> I don't mind what any of these values are, but I would strongly recommend asking @kantp to run his test with these revised values and see if it still runs as expected.

I'll do this since Europe has already reached the weekend.

src/Network/Broadcast/OutboundQueue.hs
dcoutts: Because this is the highest priority message that edge nodes send. Edge nodes do not send block header announcements, so we do not expect to find block header announcements in the queue ahead of us. Whereas for core and relay nodes, they do send block header announcements, and those take the maxahead 0 position.

src/Network/Broadcast/OutboundQueue.hs
dcoutts: Because p2p nodes *do* send block header announcements, which are higher priority than requests for blocks.

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -128,7 +140,7 @@ enumPrecHighestFirst = reverse enumPrecLowestFirst
 --
 -- If we cannot find any alternative that doesn't match requirements we simply
 -- give up on forwarding set.
-newtype MaxAhead = MaxAhead Int
+data MaxAhead = MaxAhead Int
```
dcoutts: This could go back to being a newtype, but no big deal.

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -267,28 +336,57 @@ newtype MaxInFlight = MaxInFlight Int
...
+-- | Default dequeue policy for core nodes
+defaultDequeuePolicyCore :: DequeuePolicy
+defaultDequeuePolicyCore = go
   where
-    -- Dequeueing policy for core nodes
     go :: DequeuePolicy
-    go NodeCore  = Dequeue NoRateLimiting (MaxInFlight 2)
-    go NodeRelay = Dequeue NoRateLimiting (MaxInFlight 1)
+    go NodeCore  = Dequeue NoRateLimiting (MaxInFlight 3)
+    go NodeRelay = Dequeue NoRateLimiting (MaxInFlight 2)
```
dcoutts: I'm happy to have this change for now, until we can resolve and test the re-entrant conversation business.

dcoutts:
Can squash and merge once CI is green.

deepfire:
CI is green!

@dcoutts, I think I'm going to start deploying this.

deepfire:
@dcoutts, the CI [1] on `cardano-sl` suggests that apparently we need a correspondingly updated version of `cardano-sl`?

-- 
1. https://hydra.iohk.io/build/16477/nixlog/1/tail

PR #90, opened 2017-08-15
[CSL-1505] Do not define default policies; peer classification for non-routed peers

avieth:
They've been moved to cardano-sl. ~~Will reference the pull request once it's been made.~~

https://github.com/input-output-hk/cardano-sl/pull/1347

Also included a commit to extend the `Peers` datatype with a classification of those peers which do not necessarily appear in the routing lists. This is useful because sometimes we route to peers even though they don't appear in our routing list, as in `MsgGetBlocks`.


edsko:
After discussing this with @dcoutts , the advantage of the approach in this PR is that by having this mapping from node IDs to node types as part of the `Peers` structure, we can piggy-back on the support for multiple buckets of peers and hence allow different threads to update their own mapping independently.

PR #93, opened 2017-08-17
[CSL-1515] decrement in flight before delay

avieth:
Previously, a message which had been dequeued and completed would still
occupy space in the queue. How long it remains effectively in the queue
depends upon the rate limit. In cardano-sl a rate limit of 1 per second
is used, and so it was observed that a message would always stay in the
queue for at least one second. This caused recovery mode to fail. The
MsgRequestHeaders would hang around for ~0.8 seconds after the
conversation had finished, during which time two MsgRequestBlocks are
enqueued, but the second would fail because there's 2 message ahead of
it. Failing to get the needed blocks, the node would try again,
enqueuing a MsgRequestHeaders and repeating the cycle of failure.

## Update August 18

The problem arises in recovery mode. There's a block retrieval queue and one
thread (the retrieval worker) which clears it. This is in `Pos.Block.Network.Retrieval`.
If it pulls out a header which is determined to be for an alternate chain (alternate
meaning its parent is not our tip, so not necessarily a fork, just a header which
does not *immediately* continue our chain) then it will enter recovery mode via
`triggerRecovery`.

This term `triggerRecovery` is currently the only program to enqueue a
`MsgRequestBlockHeaders`. Find it in `Pos.Block.Network.Logic`. Within
the conversation is requests the peer's tip header (`requestTip`, same module)
and then classifies it and, if it's useful, puts it back into that retrieval queue.

There was never a problem with the `MsgRequestBlockHeaders` enqueueing.
It would typically work as expected, close up for all peers (`triggerRecovery`
would wait for all conversation to finish), and return to the retrieval worker.
The story so far is that we've found a header for an alternative chain, so we're
at `handleAlternative` in `Pos.Block.Network.Retrieval`. The recovery mode
`TMVar` is updated so that the rest of the system knows that recovery mode is
on, and then a `MsgRequestBlocks` is enqueued to the peer which gave us the
relevant header (the one that we put into the recovery `TMVar`).

That `MsgRequestBlocks` doesn't actually request blocks right away. Instead it
asks for more headers, hoping to use them to find the nearest ancestor in its own
chain. TW-161 proposes to use a `MsgRequestBlockHeaders` type here. Within that
conversation, `handleCHsValid` will be called which also enqueues a
`MsgRequestBlocks`, this time actually asking for the blocks themselves, for each
of the headers, from that same peer, and waits until the conversation finishes.
This is the nested `MsgRequestBlocks` conversation that was causing trouble.
Without these changes, the `MsgRequestBlockHeaders` from earlier was still
occupying space in the queue, so the second `MsgRequestBlocks` was denied
and recovery mode was judged a failure and restarted.

@dcoutts noticed that there's a potential for even more nesting! In `handleCHsValid`,
if there's an error retrieving the blocks, `triggerRecovery` will be entered yet again
via `dropRecoveryHeaderAndRepeat`! That's to say, `MsgRequestBlockHeaders`
will be enqueued and waited on within a `MsgRequestBlocks`, within a `MsgRequestBlocks`.

### DIFF
--- a/src/Network/Broadcast/OutboundQueue.hs
+++ b/src/Network/Broadcast/OutboundQueue.hs
@@ -775,9 +775,10 @@ intEnqueue outQ@OutQ{..} msgType msg peers = fmap concat $
 
 -- | Node ID with current stats needed to pick a node from a list of alts
 data NodeWithStats nid = NodeWithStats {
-      nstatsId      :: nid  -- ^ Node ID
-    , nstatsFailure :: Bool -- ^ Recent failure?
-    , nstatsAhead   :: Int  -- ^ Number of messages ahead
+      nstatsId       :: nid  -- ^ Node ID
+    , nstatsFailure  :: Bool -- ^ Recent failure?
+    , nstatsAhead    :: Int  -- ^ Number of messages ahead
+    , nstatsInFlight :: Int  -- ^ Number of messages in flight
     }
 
 -- | Compute current node statistics
@@ -787,8 +788,8 @@ nodeWithStats :: (MonadIO m, WithLogger m)
               -> nid
               -> m (NodeWithStats nid)
 nodeWithStats outQ prec nstatsId = do
-    nstatsAhead   <- countAhead outQ nstatsId prec
-    nstatsFailure <- hasRecentFailure outQ nstatsId
+    (nstatsAhead, nstatsInFlight) <- countAhead outQ nstatsId prec
+    nstatsFailure                 <- hasRecentFailure outQ nstatsId
     return NodeWithStats{..}
 
 -- | Choose an appropriate node from a list of alternatives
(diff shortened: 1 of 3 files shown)


avieth:
@edsko @dcoutts b03670e75c33a3cab2b4a0858535a199c6efa166 is all that's needed and it's very small.

The other commit isn't necessary but I suspect it might be useful in the future so I left it in. I'll pop it off either of you insist.

avieth:
Also noticed that the rate limit applies to all target node types, but the configuration has a particular rate limit for each type of peer. So a rate limit of 1 to core nodes, for instance, can slow down traffic to relay nodes.

kantp:
I've used this branch, together with the current master of `cardano-sl` (`fd0b17f357fd561467681d086a2e4fda6c3a864e` to be precise and for future reference), and have observed recovery to succeed.

In particular, I've started a cluster of 3 core, 1 relay, and 1 edge nodes, and after some time killed and restarted the relay node. The relay entered recovery mode, and completed it gracefully. Also (because removing the relay prevented it from getting updates), the edge node went into recovery mode, and was able to complete it as well.

Furthermore, I see a delay of 1 second between enqueueing and sending of the two successive `MsgRequestBlocks` requests, so rate limiting is indeed still in place.

avieth:
> The queue does not dequeue a message to a node with an in-flight measure over a specified threshold. By decrementing the in-flight count only after a certain delay, this was effectively implementing dequeue rate limiting. What @avieth objects to in this ticket is that this also limits enqueueing; I thought that was somewhat by design (we discussed previously how the demo needed to be adjusted not to enqueue too many transactions at once), but if this is undesirable, then this PR is not quite sufficient -- by decrementing the count immediately after the send, enqueueing is indeed not rate limited, but neither is dequeueing. Instead I suppose we need to be more explicit about rate limiting. I will try to come up with a fix.

I see. I'd expect that when something is dequeued it immediately makes more room for an enqueue, but that the rate of dequeues is limited.

avieth:
> I see. I'd expect that when something is dequeued it immediately makes more room for an enqueue, but that the rate of dequeues is limited.

And as far as I understand that's exactly how it works now. LGTM

avieth:
@dcoutts I'll wait for you to give a pass.

Perhaps we should take some time to write tests for `OutboundQ`

PR #96, opened 2017-08-24
[CSL-1549] correct enqueue subset

avieth:
Peers which are not routed in the queue's peers term now appear in the
'unknown' set and are appropriately classified and routed.


avieth:
CI exceeded maximum time limit D:

avieth:
Didn't even have time to build the dependencies `stack --no-terminal --install-ghc build --only-dependencies --jobs=1`. Could we bump up that `--jobs` option?

avieth:
Anyway, force pushed to try the builds again and my fingers are crossed.

avieth:
@dcoutts nice, thanks for that travis update. I was wondering if there was a good reason for using `-j1`.

PR #97, opened 2017-08-25
Recent failure does not disqualify node

avieth:
Instead of never trying to reach a node with a recent failure, just
demote them in the ordering, so that they'll only be tried if there are
no nodes *without* recent failures and satisfying the max ahead and
in flight constraint.


dcoutts:
We decided to take a different approach and mark nodes as not failed when they talk to us.

PR #98, opened 2017-08-28
[TW-162] Fix negative values for inflight count

adinapoli-iohk:
Fix PR fixes https://issues.serokell.io/issue/TW-162.

The idea is to replace `inFlightFor` in favour of a `setInFlightFor` function which won't touch the `InFlight` map in case the requested value cannot be found.

The other two lenses have been downgraded to getters, to avoid shooting ourselves in the foot.

I have added a test to indeed replicate the bug (I had success with that) and successively checking the bug was gone. Such test piggybacks heavily on `relayDemo`, and:

* Create a bunch of nodes where everybody talk to everybody;
* Send a bunch of async messages
* Removes peer asynchronously, to simulate the situation where we have still inflight messages which need to be acknowledged but the relevant entry in the `InFlight` map is now gone.

cc @avieth @dcoutts @edsko


src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -312,14 +313,23 @@ type InFlight nid = Map nid (Map Precedence Int)
...
+-- calculated by the updater. In case no match can be found, this function is
+-- effectively a noop.
+setInFlightFor :: (MonadIO m, Ord nid)
+               => Packet msg nid a
+               -> (Int -> Int)
+               -> MVar (InFlight nid)
+               -> m ()
+setInFlightFor Packet{..} updater var = liftIO $ modifyMVar_ var $ \iflight -> do
+    let updateInner = Map.adjust updater packetPrec
+    return $ Map.adjust updateInner packetDestId iflight
```
dcoutts: Minor nitpick: this could be more functional and less imperative in style:
```
setInFlightFor Packet{..} f var =
    liftIO $ modifyMVar_ var (return . update)
  where
    update :: InFlight nid -> InFlight nid
    update = Map.adjust (Map.adjust f packetPrec) packetDestId 
```

src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -25,6 +25,7 @@
 {-# LANGUAGE ScopedTypeVariables       #-}
 {-# LANGUAGE StandaloneDeriving        #-}
 {-# LANGUAGE TupleSections             #-}
+{-# OPTIONS_GHC -fno-warn-redundant-constraints #-}
```
dcoutts: Do we really need this? We'd prefer to keep it clean and fix the redundant constraints.

src/Network/Broadcast/OutboundQueue.hs
adinapoli-iohk: ok!

src/Network/Broadcast/OutboundQueue.hs
adinapoli-iohk: Ah! This was needed because the minute I swapped a `Lens` for a `Getter` GHC complained about the unused `Contravariant` constraint, which was probably pulled in scope by the `Getter` definition. No clue how to fix that, I'd confess 😁

PR #99, opened 2017-08-31
[TW-163] Extend registerQueueMetrics to accept an optional namespace

adinapoli-iohk:
This PR extends `registerQueueMetrics` with an optional `Maybe Namespace` parameter so that we can prefix our EKG metrics with more appropriate namespaces.

This is how things will appear in EKG after merging this PR:

screen shot 2017-08-31 at 13 30 01


src/Network/Broadcast/OutboundQueue.hs
```diff
@@ -642,7 +655,18 @@ registerQueueMetrics OutQ{..} store = do
...
+        let initialList = ["queue", qSelf]
+            -- If we have a namespace, add it to the beginning of the
+            -- initial list, so that the final metric name will be
+            -- correctly namespaced.
+            -- E.g.
+            -- let namespaceMb = Just "mynamespace"
+            -- let qSelf       = "test"
+            -- metric ["InFlight"]
+            -- >>> mynamespace.queue.test.InFlight
+            tokens      = maybe initialList (: initialList) namespaceMb
```
dcoutts: simpler:

     let prefix = maybeToList namespaceMb ++ ["queue", qSelf]
      in T.pack . intercalate "." . (prefix ++)

adinapoli-iohk: believe it or not TIL about `maybeToList`, thanks! 😀  (I get constantly amazed by the sheer number of useful combinator in `base`).

adinapoli-iohk:
Alas I have spotted another module which is registering metrics without namespaces:

https://github.com/serokell/time-warp-nt/blob/40f4235cda65c746c6edc2a3455242d96d7175bf/src/Node/Util/Monitor.hs#L26-L48

My preference would be to move the entire `Node.Util.Monitor` module out of this package (if feasible), so that we could also drop the dependency on `ekg` in favour of just `ekg-core`:

https://input-output-rnd.slack.com/archives/C37V4CVU5/p1504183695000589

adinapoli-iohk:
@dcoutts I have decided to go for the kill (probably worth sharing the knowledge with @edsko and @avieth here). 

This last commit removes the `Node.Util.Monitor` from `timewarp-nt` (we will re-add it as part of cardano). My rationale can be summarised by these bullet points:

* It doesn't seem to belong to timewarp-nt anyway (aka it feels a bit like a scope creep, but of course this boils down to personal taste);

* It imposes downstream packages an unnecessary dependency on `ekg`, where such packages might decide to use different implementations of `ekg` like `ekg-wai`.

* We cannot customise the way we register the metrics. We could surely have extended `startMonitor` to have an optional, external namespace passed as an extra argument, but that felt too much work in respect to that the module offered if it was pruned by the codebase.

Let me know what you guys think!

adinapoli-iohk:
See https://github.com/input-output-hk/cardano-sl/pull/1464#issuecomment-326303269 for the end result.

adinapoli-iohk:
Unexpected twist of the plot 😁  @dcoutts after a second review thinks it's best to rework slightly `setupMonitor` but keep it inside `timewarp-nt`. Will do this as a separate PR!

adinapoli-iohk:
The `Node.Util.Monitor` module is back as part of https://github.com/serokell/time-warp-nt/pull/100 🎉

PR #100, opened 2017-08-31
Restore setupMonitor

adinapoli-iohk:
This PR restores `Node.Util.Monitor`, but only `setupMonitor`.

This allows us to still drop `ekg`.

cc @dcoutts

### DIFF
--- a/src/Node/Util/Monitor.hs
+++ b/src/Node/Util/Monitor.hs
@@ -4,7 +4,7 @@
 {-# LANGUAGE GADTs #-}
 
 module Node.Util.Monitor (
-      setupMonitor
+      registerMetrics
     ) where
 
 import Control.Monad.IO.Class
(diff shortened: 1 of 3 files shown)


src/Node/Util/Monitor.hs
```diff
@@ -28,44 +24,27 @@ setupMonitor
...
        )
-    => (forall t . m t -> IO t)
+    => Maybe T.Text
+    -> (forall t . m t -> IO t)
     -> Node m
     -> Monitoring.Store
     -> m Monitoring.Store
-setupMonitor lowerIO node store = do
-    liftIO $ flip (Monitoring.registerGauge "Remotely-initated handlers") store $ lowerIO $ do
+setupMonitor mbNamespace lowerIO node store = do
```
dcoutts: It should return `m ()` and be called `registerMetrics`

PR #105, opened 2017-09-13
add the "bucketSpareCapacity" function

adinapoli-iohk:
This simple PR adds a new function called `bucketSpareCapacity` to be used to query an `OutboundQueue` regarding the "free slots" it has to accept more peers. It can be used by application code to implement services to check if we hit our subscription limit for a particular <queue, bucket> combination.

cc @edsko @dcoutts @avieth

### DIFF
--- a/src/Network/Broadcast/OutboundQueue.hs
+++ b/src/Network/Broadcast/OutboundQueue.hs
@@ -47,6 +47,7 @@ module Network.Broadcast.OutboundQueue (
   , ReconsiderAfter(..)
     -- ** Subscription
   , MaxBucketSize(..)
+  , bucketSpareCapacity
     -- * Enqueueing
   , Origin(..)
   , EnqueueTo (..)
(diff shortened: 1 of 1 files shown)


adinapoli-iohk:
> Using 'Nothing' to mean 'As much as you like' is a bit confusing semantically though

I do agree! Any suggestion? I guess we could create a separate type like:

```haskell
data BucketCapacity = CapacityUnlimited
                                    |  BucketCapacity !Int
```

It would act essentially as a non-polymorphic version of `Maybe a`, but with a more precise semantic. I'm easy!

edsko:
It doesn't really return the capacity though, does it? It returns _spare_ capacity?

adinapoli-iohk:
> It returns spare capacity?

Correct. So I guess we either name the type appropriately (I suck at names!) or we simply stick with the status quo (`Maybe`) :)

edsko:
How about

``` haskell
data SpareCapacity = SpareCapacity Int | UnlimitedCapacity
```

:)

adinapoli-iohk:
Do like! :)

adinapoli-iohk:
@edsko ok, it should be semantically less confusing now! 😛

PR #106, opened 2017-09-20
[CSL-1652] Improve removePeer

adinapoli-iohk:
This maybe-over-complicated PR improves `removePeer` to skip empty lists.

See here for the full rational and debugging: https://issues.serokell.io/issue/CSL-1652#comment=96-7493

I have also added a bunch of properties to demonstrate this was indeed an issue prior to this patch:

screen shot 2017-09-20 at 16 56 24

But that now everything is green:

screen shot 2017-09-20 at 17 24 53

### DIFF
--- a/src/Network/Broadcast/OutboundQueue/Types.hs
+++ b/src/Network/Broadcast/OutboundQueue/Types.hs
@@ -233,10 +233,7 @@ removePeer toRemove peers =
     -- Removes the peer `toRemove` from the list of alternatives, skipping
     -- any empty list encountered, which is discarded as meaningless information.
     remove :: AllOf (Alts nid) -> AllOf (Alts nid)
-    remove [] = []
-    remove (x:xs) = case filter (/= toRemove) x of
-        [] -> remove xs
-        l  -> l : remove xs
+    remove = filter (not . null) . map (filter (/= toRemove))
 
 {-------------------------------------------------------------------------------
   Classification of messages and destinations
--- a/src/Network/Broadcast/OutboundQueue/Types.hs
+++ b/src/Network/Broadcast/OutboundQueue/Types.hs
@@ -32,6 +32,7 @@ import Data.Set (Set)
 import qualified Data.Set as Set
 import Data.Maybe (mapMaybe)
 import Control.Lens
+import Test.QuickCheck as QC
 import Formatting
 
 -- | Node types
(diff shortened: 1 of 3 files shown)


src/Network/Broadcast/OutboundQueue/Types.hs
```diff
@@ -219,8 +230,13 @@ removePeer toRemove peers =
...
             }
   where
+    -- Removes the peer `toRemove` from the list of alternatives, skipping
+    -- any empty list encountered, which is discarded as meaningless information.
     remove :: AllOf (Alts nid) -> AllOf (Alts nid)
-    remove = map $ filter (/= toRemove)
+    remove [] = []
+    remove (x:xs) = case filter (/= toRemove) x of
+        [] -> remove xs
+        l  -> l : remove xs
```
edsko: how about just 

``` haskell
remove = filter (not . null) . map (filter (/= toRemove))
```

Not essential though obviously :)

adinapoli-iohk: Genuine question for you @edsko : do you think that doing the `filter + map + filter` combo will trigger fusion rules or are we effectively traversing the list 3 times? You know, that one liner was indeed what I jumped at immediately, but couldn't find enough evidence in the GHC docs that performance wasn't going to suffer too much. I know that our forwarding sets it's going to be very small anyway and this probably is premature optimisation, but I guess it's good to know for future references :)

edsko: Fusion isn't so relevant in a lazy language. Sure, it's nice when possible to avoid unnecessary memory allocation, but we're not actually traversing the list two times. Also, in the vast majority of cases, code clarity trumps minor local optimizations like this.

adinapoli-iohk: Perfect. In which case, I'm inclined to go with the one liner even more 😉

edsko:
Did my version pass your QuickCheck tests? :)

adinapoli-iohk:
> Did my version pass your QuickCheck tests? :)

Of course it did! It would be imprudent, as much as I respect you guys, to commit untested code 😜

edsko:
> Of course it did! It would be imprudent, as much as I respect you guys, to commit untested code :stuck_out_tongue_winking_eye:

I'm not sure whether to be impressed or insulted :p

adinapoli-iohk:
> I'm not sure whether to be impressed or insulted :p

Maybe a bit of both? :trollface: 😬 😹

adinapoli-iohk:
Thanks @gromakovsky !

PR #108, opened 2017-09-28
[CSL-1709] Log the reason why the connection was lost.

adinapoli-iohk:
This tiny PR stems from a conversation with @avieth and from [CSL-1709](https://issues.serokell.io/issue/CSL-1709). In order to troubleshoot and debug some subtle issues, it's nice to get as many logging as we can of the underlying error causes.

In particular, this PR logs the reason why a connection was lost. Hopefully it will be enough to track down the place inside `network-transport-tcp` where this is coming from (and why), to understand better what's going on.

Feel free to add more useful logging as part of this PR.

### DIFF
--- a/src/Node/Internal.hs
+++ b/src/Node/Internal.hs
@@ -835,7 +835,8 @@ nodeDispatcher node handlerInOut =
 
           -- When a heavyweight connection is lost we must close up all of the
           -- lightweight connections which it carried.
-          NT.ErrorEvent (NT.TransportError (NT.EventErrorCode (NT.EventConnectionLost peer bundle)) _msg) ->
+          NT.ErrorEvent (NT.TransportError (NT.EventErrorCode (NT.EventConnectionLost peer bundle)) reason) -> do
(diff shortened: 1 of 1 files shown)


avieth:
LGTM obviously but I'll take a little time to think about what else we might want to log, for CSL-1709.

avieth:
I suspect we're going to see that it's a socket exception (premature exit, rather than the peer closing the connection or closing its own endpoint). In the CSL-1709 case, the edge node must successfully make a connection and receive confirmation from the peer (nt-tcp handshake), else the connection wouldn't even be brought up and no connection lost event would be posted for it.
The next thing it will do is try to send its own peer data and then send a `SYN` code. Then peer is expected to send its peer data followed by the `ACK`. It'd be good to know at exactly what point it fails. Right now all we know is that it's some time before the remote peer data is received and parsed by the edge node.
The failure could be on either side.

It might be useful to log every network-transport event other than receive (since that would be so much noise). Maybe a `ConnectionOpened` event comes through before the connection is lost?

PR #111, opened 2017-10-19
Introduce compatibility with log-warper 1.4

volhovm:
cc @ChShersh


avieth:
@ChShersh @volhovm what's the status of this. Perhaps we should update it to use log-waper 1.7.2?

volhovm:
Oh my god, it's still alive. I will bump compatibility after time-warp is merged into CSL, alright? 
Yes, it should be 1.7.2 in csl master.

PR #113, opened 2017-11-06
[CSL-1804] Various sprint planning markdowns

avieth:


notes/shelley/Interface.md
```diff
@@ -0,0 +1,461 @@
...
+Sometimes the block retrieval worker will trigger recovery mode, in case it
+finds a header which doesn't immediately continue the tip. This recovery
+process uses time-warp conversations for control flow: it will get the headers
+and blocks needed to go from current tip to new tip, and then return to the
+block retrieval queue clearing loop, possibly ending recovery mode.
+
+We can keep this as is (equally bad, not worse) by using the synchronous
+`getBlocks` function and letting the diffusion layer take care of finding them.
+
+## Concrete details of implementation
```
avieth: From here down is some actual Haskell text of how I think we should form this. There's also `Interface.hs` which shows a working example without any actual functionality. I'm ready to run with this but obviously I'd like some feedback first.

@dcoutts @georgeee

notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+  choose a bunch of peers based on some XOR, then gather some data about their
+  delta-Q, then choose a subset according to percentiles of delta-Q/performance.
+  Difference here is that the computation of delta-Q is wholistic: we're
+  concerened with the entire path (which can be known here) rather than in the
+  non-core-DIF where we're concerned only with the first hop.
+
+- Some discussion about time constraints RE: ensuring that the slot leader's
+  block goes out earlier enough that it reaches the next slot leader before
+  it's too late.
+  Always going for worst case actually lowers best case throughput (i.e. tps).
```
dcoutts: I think we concluded we should send out the block at the start of the slot, and not wait.

notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+  throughout the network.
+
+- Matrix of delta-Q's (100 x 100). How to fill it up? We know our own, we
+  propagate our own throughout the network (must propagate in order to
+  propagate, hm...)
+  I'm about to publish. I look at the matrix and I can compute paths which
+  with some certainty will get the data to everybody within some time. Of
+  course, it depends upon what the forwarders do.
+  Minimize maximum delay.
+  Also, due to paranoia, send to a few others with some random choice.
```
dcoutts: We need to clarify what this algorithm is with Neil.

notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+performance. It's hard enough for a single computer to make time guarantees,
+let alone a network of them.
+
+## Questions after meeting with Neil and Duncan November 8
+
+1. Nodes in the core DIF never hold coins? Only stake? Why? Does this buy any
+   security at all? I suppose the machine itself is a target, and you wouldn't
+   want to keep the private key to the _coins_ on that machine. Keeping the
+   key to the delegate is fine, because the holder of the coins can revoke if
+   there's a compromise. Correct?
```
dcoutts: Yes, only stake keys on these servers, and we'll almost certainly have an online/offline stake key solution so that nodes really use an online key and present a certificate that shows that it's a valid proxy for the known stake key.

notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+
+2. Ability to deny being delegated to? Is this a must have? I would hope so.
```
dcoutts: Yes, but not for this workstream I think.

notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+
+3. What does sharing "on the blockchain" really mean? How does it work? I'm
+   very skeptical. Why not flood the (non-core-dif) network with your bearer
+   address encrypted with the target's public key?
```
dcoutts: > What does sharing "on the blockchain" really mean? How does it work?

Its much like sending out an MPC message which gets incorporated into the blockchain. So it's just a special kind of blockchain entry.

> Why not flood the (non-core-dif) network with your bearer address encrypted with the target's public key?

It means we have evidence of non-compliance. We could do it by broadcast but then we have no objective evidence later that someone didn't do what they were supposed to. Perhaps for v1 that is ok since we have problems gathering evidence for any of the other bits either.

notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+
+4. Is the core dif construction mechanism essentially the same as the bigger
+   network? Start with some XOR kademlia-style thing, then use delta-Q to
+   refine it. Only difference is the wholistic approach, in which the entire
+   weighted graph is known?
```
dcoutts: I'm not aware of anything contradicting that. And if we can make them as similar as possible that is obviously good, we can reuse logic & code.

notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+
+5. It was taken for granted that a node will know all of the pairwise delta-Q
+   of the ~100 nodes in the core DIF. How? Surely we can't expect a consistent
+   view of this. What to do in case of missing information? I'm skeptical that
+   it will just work.
```
dcoutts: This is part of the propagation algorithm that is yet to be fully defined. And it doesn't need to be fully consistent.

notes/shelley/CoreDIF.md
```diff
@@ -0,0 +1,153 @@
...
+
+6. Suppose a core dif member uses 2 relay bridges. Is it OK to give the
+   address of one of these to some nodes, and the other to some other nodes,
+   w.r.t. the delta-Q measurements? I suspect so. Some will calculate the
+   delta-Q for one relay, some for the other, but it's all effectively the
+   delta-Q to the node behind the bridge relay.
+   I don't think 1 bearer address per public key is necessary. Indeed Neil
+   was quite excited about it *not* being necessary; you can give different
+   addresses to different peers.
```
dcoutts: I think that's fine. The delta-Q is about the route from A to B.

notes/shelley/DeltaQ.md
```diff
@@ -0,0 +1,86 @@
+# QoS / delta-Q estimation
+
+Peer discovery gives us a pool of peers but no information about link quality.
```
dcoutts: Except the first & last times we exchanged UDP packets (& we could if we want maintain more, e.g. round-trip timing).

notes/shelley/DeltaQ.md
```diff
@@ -0,0 +1,86 @@
...
+To inform a choice of peers to actually communicate with, we'll need to gather
+data for an estimation of the delta-Q at a given peer.
+
+The esimate must be exposed (read-only) to the user of network-transport-tcp.
+It's not determined what exactly that data should be. The distribution itself?
+The s, g, and v components?
```
dcoutts: To check with Neil. I'd guess the latter is most useful.

notes/shelley/DeltaQ.md
```diff
@@ -0,0 +1,86 @@
...
+getMeasurements :: Transport -> EndPointAddress -> IO DeltaQStuff
+```
+
+## Implementation
+
+Must be implemented close to the socket itself, i.e. in network-transport-tcp.
+It will rely on precise time measurements on windows of data. New control
+codes would be defined in order to delimit these windows, i.e. indicate a
+leading and trailing edge. This allows for an estimation of the s component
+of delta-Q (latency as a function of the amount of data sent).
```
dcoutts: We already preserve message boundaries on lightweight channels.

notes/shelley/DeltaQ.md
```diff
@@ -0,0 +1,86 @@
...
+  - When trailing edge was sent
+  - When leading edge was received remotely (different clock, possibly lying)
+  - When trailing edge was received remotely (different clock, possibly lying)
+  - When leading edge was received locally
+  - When trailing edge was received locally
+
+To get a precise timestamp for a receive, we can use the SIOCGSTAMP ioctl.
+This is the timestamp of the last packet delivered to userspace. We would call
+it after receiving a leading or trailing edge, and before doing the next
+`recv`.
```
dcoutts: It may be tricky to work out how these match up with the message boundaries. We'll have to read the docs on what this ioclt means exactly for tcp connections.

notes/shelley/DeltaQ.md
```diff
@@ -0,0 +1,86 @@
...
+get the TCP round-trip-time estimate. Are we interested in that?
+
+## Relationship to input QDisc
+
+network-transport-tcp endpoints may have nontrivial queueing policies. Should
+latency induced by queueing factor into the delta-Q estimate? That's to say,
+are we trying to estimate only TCP round-trip-time, or network-transport-tcp
+round-trip-time? Neither option poses any extra challenge, it's only a matter
+of deciding when to respond to the control message: before or after the
+enqueue.
```
dcoutts: Good Q. Another point to discuss with Neil.

notes/shelley/Interface.hs
```diff
@@ -0,0 +1,83 @@
...
+
+data LogicLayer = LogicLayer
+  { logic :: Logic
+  , runLogicLayer :: IO ()
+  }
+
+data DiffusionLayer = DiffusionLayer
+  { diffusion :: Diffusion
+  , runDiffusionLayer :: IO ()
+  }
```
dcoutts: Lets document for each of these what it is exactly. For example "the interface provided by the logic layer to the diffusion layer, for the diffusion layer to make calls into the logic layer".

notes/shelley/Interface.md
```diff
@@ -0,0 +1,525 @@
...
+
+A good first step in Shelley networking implementation is to carve out the
+interface between application and diffusion, so that we can do our work
+independently (besides the architectural advantages).
+
+In summary: all "conversation" and relaying is done by the diffusion layer
+only. This way application latency does not inflate network latency. Input to
+the application layer will come from the diffusion layer rather than from a
+set of network listeners. Output to the network will go via the diffusion
+layer, and may not offer any feedback on success or failure.
```
dcoutts: For emitting transactions we probably do still need some notification of failure because the wallet needs some retry logic.

notes/shelley/Interface.md
```diff
@@ -0,0 +1,525 @@
...
+--
+-- A 'RawBlockHeader' is given, i.e. it's decoded from the wire but not
+-- validated. Application layer can validate it as needed.
+getTips :: IO [RawBlockHeader]
+
+-- Offer a block for diffusion.
+-- Is it OK to return right away? Or should the diffusion layer respond with
+-- some indication of whether the block was actually sent to some threshold of
+-- peers?
+announceBlock :: Block -> IO ()
```
dcoutts: Depends on whether the logic layer needs to act on it (beyond logging). If it is just logging, that can be done in the diffusion layer.

notes/shelley/Interface.md
```diff
@@ -0,0 +1,525 @@
...
+
+-- Similar for other data that can be broadcast
+-- (transactions, ssc, update, etc.)
```
dcoutts: But for transactions I think we do need a failure indication.

notes/shelley/Interface.md
```diff
@@ -0,0 +1,525 @@
...
+```
+
+```Haskell
+-- Direction: diffusion => application
+
+-- Give various data to the application layer asynchronously (drop it in some
+-- queue).
+block :: Block -> IO ()
```
dcoutts: Do we notify of blocks like this? I thought it was only headers, and they logic layer has to call `getBlock` or `getBlocks` to get it/them.

notes/shelley/CoreDIF.md
georgeee: What kind of evidence are you talking about?

Anyway, also vote for broadcasting.
When one wanna join Core DIF, he broadcasts message he wants to get addresses of DIF relay nodes and retrieves these encrypted in some way. Then through DIF relay nodes he can authenticate to finally join DIF. See no need, room or rationale to do it on chain.

notes/shelley/BlockRetrievalRecovery.md
```diff
@@ -0,0 +1,126 @@
...
+
+This is fine, all we're doing is changing the source of the block headers.
+Instead of coming from an explicit time-warp conversation, they'll come
+from the diffusion layer (which uses a time-warp conversation, but that's
+abstracted). We can get rid of `handleUnsolicitedHeaders` because all
+we need is `handleUnsolicitedHeader`, which can be called whenever the
+diffusion layer presents a block header. `handleUnsolicitedHeader` requires
+a `NodeId` of the provenance of the block, but that's only there so it
+knows who to ask for the body, so it should be removed because it's the
+diffusion layer's responsibility to know where to find the body.
```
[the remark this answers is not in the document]
avieth: Yes I imagine that's how it would work. Included in the cache would be a set of `NodeId`s for peers which have announced that header.

notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,140 @@
...
+It's enlightening to look at the proposed interface presented by the diffusion
+layer (where pub/sub lives) and the application layer (see this
+[document](./Interface.md)). It's transparent to pub/sub; the application
+layer won't know about it. Functions like `announceBlock` which diffuse data
+will induce a publish of that thing. Every subscriber for the block topic will
+be delivered the block.
+
+Functions like `getBlocks` which bring data up from the diffusion layer may
+need to request that data from peers, so a hybrid request/response and pub/sub
+model is needed.
```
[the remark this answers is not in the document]
avieth: I'll extend this with the notes I have from our meeting on Friday.

notes/shelley/CommunicationPatterns.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Number of subscribers
+
+Each subscription will be served by one TCP connection, so the OS imposes a
+limit on the number of subscribers. It's plausible that even far below this
+limit performance will start to suffer. But it's also plausible that the number
+of subscriptions actually needed per node is low enough (not more than a
+hundred) to avoid any such problems.
+There's already a maximum subscribers configuration in the outbound queue that
+can be reused.
```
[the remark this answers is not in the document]
avieth: https://github.com/haskell-distributed/network-transport/issues/25
https://github.com/haskell-distributed/network-transport/issues/17

notes/shelley/CommunicationPatterns.md
```diff
@@ -21,21 +22,75 @@ to it for that same topic. Call it lazy subscription.
...
+      This is dodgy. It means the publisher failed to deliver before TTL.
+    If in state 3, forget `k` and `v` and enter state 1.
+
+### Stretch: privacy considerations
+
+There are known deanonymization schemes effective against cardano (and other
+cryptocurrencies including bitcoin) which exploit predictable patterns of
+broadcast. It's not a requirement for Shelley, but eventually we may want to
+use a mitigating broadcast scheme. Such a solution could include bouncing
+new transactions through other edge nodes, by way of an intermediate core node.
```
coot: Do you foresee something like cardano on top of tor-like network?

notes/shelley/CommunicationPatterns.md
```diff
@@ -95,6 +150,79 @@ and is missing some that we will need.
...
+      detecting half-open TCP connections (intermittent lower-layer failures).
+
+### Publisher sends
+
+  - Announce key : indicate that a payload for this key is available.
+  - Payload value : give the payload for a key.
+
+Should the publisher acknowledge add/remove topic?
+If a subscriber adds a topic and then removes it, there may be announces for
+that topic inbound.
```
coot: We could send them back with the ack message for remove, or if we don't want them, remove the subscriber only after the queue was cleared up to remove request.  If we have all the requests in a FIFO this property will hold.

PR #114, opened 2017-11-21
Node can wait for dispatcher thread

avieth:
Useful in case you want to run a server "forever" i.e. until it crashes.


avieth:
Not necessary. Closing

